PR #244, opened 2021-06-14
feat(openldap): Add OpenLDAP role

Ferenc-:


roles/openldap/defaults/main.yml
```diff
@@ -0,0 +1,194 @@
...
+docker_registry_openldap: quay.io
+
+##########################
+# openldap
+
+openldap_configuration_directory: /opt/openldap/configuration
+openldap_secrets_directory: /opt/openldap/secrets
+openldap_docker_compose_directory: /opt/openldap
+
+openldap_network: 172.31.100.32/28 # TODO: Find a proper one not shared with phpmyadmin
```
berendt, line 20: Check http://docs.osism.de/appendix/tables/service-networks.html. There is a table with all used network. The last assigned one is ``172.31.100.224/28``. The next free one is ``172.31.100.240/28``.

roles/openldap/defaults/main.yml
```diff
@@ -0,0 +1,194 @@
...
+openldap_container_name: openldap
+
+openldap_host: 192.168.16.5 # TODO: Double check this, perhaps use 127.0.0.1
```
berendt, line 23: Yes. Use ``127.0.0.1`` as default here.

Ferenc-: Done

roles/openldap/defaults/main.yml
Ferenc-: Done

berendt:
@Ferenc- Sorry for the delay. I hope to get to it by Monday.

roles/openldap/defaults/main.yml
Ferenc-: So the way I understood, these settings here are overridden by the settings in 
[ansible-defaults](https://github.com/osism/ansible-defaults/blob/main/all/099-infrastructure.yml#L21).
But what I so far experienced is that this value somehow still remains on `127.0.0.1`, and doesn't take the `192.168.16.5` value that I would expect in pluscloud.
I tried to follow the keycloak way, but I can't find the difference, yet keycloak gets the right address  `192.168.16.5`.
Do you perhaps have a tip what might be still missing?

Ferenc-: Never mind, it must have been a transient problem. Now it's gone.

PR #245, opened 2021-06-22
feature redhat for role chrony

muellerbe:


berendt:
Molecule tests will be solved later.

PR #304, opened 2021-09-30
docker: do not restart the service on manager nodes by default

### LINKED ISSUE #303: docker_allow_restart breaks osism-generic bootstrap
since the bootstrap allows the docker daemon to restart after creating /etc/docker/daemon.json on the manager itself, the deployment breaks because the manager_osism-ansible_1 container will be also restarted and the bootstrap kills itself.

What was the reason to restart the docker daemon in that place? from documentation after bootstrap there is a planned reboot, so from my point of view there is no need to restart the docker here.

https://github.com/osism/ansible-collection-services/blob/a06ff300d36f10cd3825908832b7ee8fe6d75912/roles/docker/handlers/main.yml#L2-L7

berendt:
Closes #303

### DIFF
--- a/roles/docker/defaults/main.yml
+++ b/roles/docker/defaults/main.yml
@@ -7,7 +7,12 @@ docker_experimental: "false"
 docker_live_restore: "true"
 docker_user: "{{ operator_user | default('dragon') }}"
 
+##########################
+# restart
+
 docker_allow_restart: true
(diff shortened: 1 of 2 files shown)


berendt:
Good point. This should fix that. You then have to set docker_enforce_restart to True in order to restart Docker on the manager.

PR #415, opened 2021-12-01
using node_id needs netaddr pip package installed

muellerbe:


roles/manager/templates/wrapper/osism-update-manager.j2
```diff
@@ -22,7 +22,7 @@ if [[ ! -x "$(command -v ansible-playbook)" ]]; then
 
         virtualenv -p python3 $VENV_PATH
         source $VENV_PATH/bin/activate
-        pip3 install --no-cache-dir "$ANSIBLE_VERSION"
+        pip3 install --no-cache-dir "$ANSIBLE_VERSION" netaddr
```
berendt, line 25: Use ``netaddr==0.8.0``

muellerbe: used NETADDR_VERSION for this, is this OK?

PR #488, opened 2021-12-17
traefik: add keepalived support

berendt:


berendt:
By switching to Kubernetes on the Mangagement Plane, this is no longer necessary.

PR #527, opened 2022-01-10
feature/add ansible tag osism filebeat

ra-rau:
- Add role filebeat
- ansible: add osism tag


berendt:
Add this one to the #486 PR.

PR #528, opened 2022-01-10
feature/add ansible tag osism metricbeat

ra-rau:
- Add metricbeat role
- ansible: add osism tag


berendt:
Please add this change to the #485 PR

PR #568, opened 2022-01-26
keycloak: add traefik integration

### LINKED ISSUE #567: keycloak: add Traefik integration

berendt:
To use Keycloak behind Traefik, a customized image of
Keycloak is required. This is necessary because the used
path keycloak must be set explicitly in various static
configuration files in the image.

The customized image is maintained in osism/container-images.

Closes #567

### DIFF
--- a/roles/keycloak/templates/docker-compose.yml.j2
+++ b/roles/keycloak/templates/docker-compose.yml.j2
@@ -4,7 +4,11 @@ version: '3.5'
 services:
   keycloak:
     container_name: "{{ keycloak_container_name }}"
+{% if not keycloak_traefik|bool %}
     image: "{{ keycloak_image }}"
+{% else %}
+    image: "{{ keycloak_for_traefik_image }}"
+{% endif %}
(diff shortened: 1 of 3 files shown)


roles/keycloak/templates/docker-compose.yml.j2
```diff
@@ -4,16 +4,46 @@ version: '3.5'
...
       - "{{ keycloak_configuration_directory }}/keycloak.env"
     volumes:
       - /etc/hosts:/etc/hosts:ro
       - /etc/localtime:/etc/localtime:ro
       - /etc/timezone:/etc/timezone:ro
+{% if not keycloak_traefik|bool %}
     ports:
       - "{{ keycloak_host }}:{{ keycloak_port }}:8080"
+{% endif %}
+{% if keycloak_traefik|bool %}
```
osfrickler, line 23: merge the "endif/if" into a simple "else"?

osfrickler (approved):

PR #570, opened 2022-01-26
traefik: do not leak the private keys

berendt:


osfrickler:
Let's still do this now, seems easy enough

PR #584, opened 2022-02-04
Add target: _blank to all homer links

osfrickler:
Some were still missing.


tibeer (approved):
LGTM!

Issue #580, opened 2022-01-28
manager: ara-server with traefik not working like expected


berendt:
The static paths are wrong.

```
  <link rel="shortcut icon" href="/static/images/favicon.ico">
  
  <link rel="stylesheet" href="/static/css/ara.css">
  
  <link id="bootstrap-dark-css" rel="stylesheet" href="/static/css/bootstrap-darkly.min.css" disabled>
  
  <link id="pygments-dark-css" rel="stylesheet" href="/static/css/pygments-dark.css" disabled>
  
  <link id="bootstrap-light-css" rel="stylesheet" href="/static/css/bootstrap-flatly.min.css">
  
  <link id="pygments-light-css" rel="stylesheet" href="/static/css/pygments-light.css">
```

berendt:
The /static problem is solvable by using STATIC_URL. But looks like it is not possible to define a prefix for the /api and other URLs.

This means this issue can only solved by switching to a host based approach.

PR #611, opened 2022-03-04
manager: support new unit based ara-server image

berendt:
Partial #580


berendt:
Will not be followed up. Needs too much switching with too little benefit.

PR #619, opened 2022-03-08
Change path prefix to host

ra-rau:
- Changed the path prefixes to host
- refactor the traefik.http....loadbalancer variable to ...-secure.loadbalancer
- This closes osism/issues#114

### DIFF
--- a/roles/cgit/templates/docker-compose.yml.j2
+++ b/roles/cgit/templates/docker-compose.yml.j2
@@ -20,22 +20,14 @@ services:
     labels:
       - "traefik.enable=true"
       - "traefik.docker.network={{ traefik_external_network_name }}"
-      - "traefik.http.services.cgit.loadbalancer.server.port=80"
       - "traefik.http.middlewares.cgit-redirect-to-https.redirectscheme.scheme=https"
       - "traefik.http.routers.cgit.entrypoints=http"
-{% if cgit_traefik_host|length > 0 %}
-      - "traefik.http.routers.cgit.rule=Host(`{{ cgit_traefik_host }}`) && PathPrefix(`/{{ cgit_traefik_path_prefix }}/`)"
-{% else %}
-      - "traefik.http.routers.cgit.rule=PathPrefix(`/{{ cgit_traefik_path_prefix }}/`)"
-{% endif %}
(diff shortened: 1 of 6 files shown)


roles/cgit/templates/docker-compose.yml.j2
```diff
@@ -20,22 +20,14 @@ services:
...
       - "traefik.docker.network={{ traefik_external_network_name }}"
-      - "traefik.http.services.cgit.loadbalancer.server.port=80"
       - "traefik.http.middlewares.cgit-redirect-to-https.redirectscheme.scheme=https"
       - "traefik.http.routers.cgit.entrypoints=http"
-{% if cgit_traefik_host|length > 0 %}
-      - "traefik.http.routers.cgit.rule=Host(`{{ cgit_traefik_host }}`) && PathPrefix(`/{{ cgit_traefik_path_prefix }}/`)"
-{% else %}
-      - "traefik.http.routers.cgit.rule=PathPrefix(`/{{ cgit_traefik_path_prefix }}/`)"
-{% endif %}
+      - "traefik.http.routers.cgit.rule=Host(`{{ cgit_traefik_host }}`)"
```
osfrickler, line 25: I wonder whether we could simply reuse the cgit_host var here and similarly for all other services. Is there a reason why we would need another variable?

roles/keycloak/templates/docker-compose.yml.j2
```diff
@@ -23,22 +23,14 @@ services:
     labels:
```
osfrickler, line 23: Github doesn't seem to allow to add comments to code outside of the change range.

In line 7-11 we are using a custom image for the traefik case, I think this is no longer needed now?

roles/netbox/templates/docker-compose.yml.j2
```diff
@@ -54,24 +54,13 @@ services:
       - "traefik.enable=true"
```
osfrickler, line 54: I'm not sure where netbox is being configured to use the expanded path that is being used in the healthcheck above, but likely that would also need to be amended?

roles/nexus/templates/docker-compose.yml.j2
```diff
@@ -36,22 +36,14 @@ services:
     labels:
```
osfrickler, line 36: Same for the healthcheck here

PR #631, opened 2022-03-15
Homer: Make unused dashboards unvisible

ra-rau:
- make that unused service dashboards not be visible
- This closes issues#546

### DIFF
--- a/roles/homer/templates/config.yml.j2
+++ b/roles/homer/templates/config.yml.j2
@@ -49,94 +49,127 @@ links:
     target: "_blank"
 
 # Services
+{% if homer_url_ara|length > 0 or homer_url_flower|length > 0 or homer_url_netbox|length > 0 or homer_url_vault|length > 0 %}
 services:
   - name: "Deployment"
     icon: "fas fa-wrench"
     items:
+      {% if homer_url_ara|length > 0 %}
       - name: "ARA Records Ansible"
         logo: "ara.png"
         subtitle: "Ansible logs"
         tag: "dashboard"
         url: "{{ homer_url_ara }}"
         target: "_blank"
+      {% endif %}
+      {% if homer_url_flower|length > 0 %}
       - name: "Flower"
         logo: "flower.png"
         subtitle: "Distributed Task Queue"
         tag: "dashboard"
         url: "{{ homer_url_flower }}"
         target: "_blank"
+      {% endif %}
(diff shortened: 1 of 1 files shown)


roles/homer/templates/config.yml.j2
```diff
@@ -52,97 +52,141 @@ links:
 services:
   - name: "Deployment"
     icon: "fas fa-wrench"
+    {% if homer_url_ara or homer_url_flower or homer_url_netbox or homer_url_vault %}
     items:
+      {% if homer_url_ara is defined and homer_url_ara|length %}
```
osfrickler, line 57: Since the URL variables are all set by default, I think it is safe to assume that they are defined, so it should be enough just to check the length. Likely an explicit "length > 0" check might be safer, though.

roles/homer/templates/config.yml.j2
```diff
@@ -52,97 +52,141 @@ links:
 services:
   - name: "Deployment"
     icon: "fas fa-wrench"
+    {% if homer_url_ara or homer_url_flower or homer_url_netbox or homer_url_vault %}
```
osfrickler, line 55: Maybe wrap the whole "- name: Deployment" block in this if? That would get rid of the "else" block.

Also, shouldn't these checks also be "| length > 0" each?

roles/homer/templates/config.yml.j2
```diff
@@ -52,97 +52,141 @@ links:
...
       - name: "Horizon"
         logo: "openstack.png"
         subtitle: "OpenStack dashboard"
         tag: "dashboard"
         url: "{{ homer_url_horizon }}"
         target: "_blank"
+      {% endif %}
+    {% else %}
+    items: []
+    {% endif %}
```
osfrickler, line 192: Missing newline

roles/homer/templates/config.yml.j2
```diff
@@ -52,97 +52,141 @@ links:
...
   - name: "User interfaces"
     icon: "fas fa-user"
+    {% if homer_url_horizon %}
     items:
+      {% if homer_url_horizon is defined and homer_url_horizon|length %}
```
osfrickler, line 182: This seems redundant.

osfrickler (commented):
"invisible" would sound better

osfrickler (approved):

PR #661, opened 2022-04-07
docker: move become to the top

berendt:
We have the become in the other roles above the task and not
below. Hence the adjustment here.

### DIFF
--- a/roles/docker/tasks/bootstrap.yml
+++ b/roles/docker/tasks/bootstrap.yml
@@ -3,11 +3,11 @@
   meta: flush_handlers
 
 - name: Add user to docker group
+  become: true
   ansible.builtin.user:
     name: "{{ docker_user }}"
     groups: docker
     append: true
-  become: true
 
 # NOTE: use paramiko connection to force a re-login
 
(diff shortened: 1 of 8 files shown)


matfechner (approved):

osfrickler (commented):
Do we have a style guide about this or could this even be enforced by linting? To me this change would only make sense if we could ensure that it is applied consistently, otherwise IMO it just clutters the history of the code.

Also, when I was writing code, I so far would have preferred the old variant here, what is the rationale to prefer the new one? To me, `become` is a modifier that should come after the action it is applied to.

berendt:
You can see better which tasks have a become when you read through a list of tasks. Since it is always in the same place between name and the actual task. I've always found it easier for me that way in the past, at least.

In most of the tasks we have, it is already implemented this way.

We can also push it always completely to the end. It would only be important to me that it is always in the same place.

Style guide for Ansible we do not have yet. So far, only implicitly via Ansible Lint. However, this does not check the become, for example. I would suggest doing that post. That we check if it is possible to add custom rules in Ansible Lint. And if not, solve it with a separate script.

I find a whole style guide a bit much. How about a contributor guide in which we then also include examples for commit messages etc. pp?

PR #680, opened 2022-05-02
import custom certs in nexus truststore

muellerbe:
Closes https://github.com/osism/issues/issues/206

### DIFF
--- a/roles/nexus/defaults/main.yml
+++ b/roles/nexus/defaults/main.yml
@@ -64,6 +64,8 @@ nexus_proxy_exclude_hosts:
   - "localhost"
   - "127.*"
   - "[::1]"
+nexus_security_ssl_url: "/service/rest/v1/security/ssl"
+nexus_security_truststore_url: "/service/rest/v1/security/ssl/truststore"
 
 nexus_docker_bearer_token_realm: true
 
--- a/roles/nexus/tasks/import-certs.yml
+++ b/roles/nexus/tasks/import-certs.yml
@@ -5,7 +5,7 @@
     url: "{{ nexus_api_url }}{{ nexus_security_ssl_url }}?host={{ item }}&port=443"
     user: "{{ nexus_admin_username }}"
     password: "{{ nexus_admin_password }}"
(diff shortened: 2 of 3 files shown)


berendt:
Please adjust the title of the PR and the commit message: https://docs.osism.tech/development/commit-messages.html#references-to-issues

roles/nexus/defaults/main.yml
```diff
@@ -64,6 +64,8 @@ nexus_proxy_exclude_hosts:
   - "localhost"
   - "127.*"
   - "[::1]"
+nexus_security_ssl_url: "/service/rest/v1/security/ssl"
+nexus_security_truststore_url: "/service/rest/v1/security/ssl/truststore"
```
osfrickler, line 68: In which scenario would these need to be changed?

Or phrased differently: Do these really need to be role variables?

roles/nexus/tasks/import-certs.yml
```diff
@@ -0,0 +1,27 @@
+---
+- name: Get PEM of URL
+  ansible.builtin.uri:
+    method: GET
+    url: "{{ nexus_api_url }}{{ nexus_security_ssl_url }}?host={{ item }}&port=443"
+    user: "{{ nexus_admin_username }}"
+    password: "{{ nexus_admin_password }}"
+    force_basic_auth: True
```
osfrickler, line 8: Please address the yaml warnings.

osfrickler (changes requested):

roles/nexus/defaults/main.yml
muellerbe: do you mean, it could be hardcoded?

roles/nexus/tasks/import-certs.yml
muellerbe: done, thanks

roles/nexus/defaults/main.yml
muellerbe: and I do not know, if there will be scenarios

PR #705, opened 2022-05-20
Update zuul and nodepool to 6.0.0

osfrickler:


osfrickler:
Self-merging as this is already running for a week.

PR #726, opened 2022-05-31
nexus: fix docker health check

grawert:
The health check needs to use the /nexus/ path to recieve a successful
http return code.

Signed-off-by: Uwe Grawert <grawert@b1-systems.de>


grawert:
The nexus/ path is valid. That is where the nexus app is reachable in the jetty container. If omited, 404 is returned. This independent from the traefik routing.

PR #736, opened 2022-06-01
Journald: Add documentation for the role

ra-rau:
Part of osism/issues#185

### DIFF
--- a/roles/journald/README.rst
+++ b/roles/journald/README.rst
@@ -1,0 +1,38 @@
+This role configures options for journald.
+
+**Role Variables**
+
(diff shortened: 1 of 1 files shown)


roles/journald/README.rst
```diff
@@ -0,0 +1,38 @@
+This role configures options for Journald.
```
berendt, line 1: Use ``journald``everywhere.

berendt (changes requested):

roles/journald/README.rst
ra-rau: done

berendt (approved):

PR #786, opened 2022-06-29
made container image names identical to other repositories

tibeer:
Signed-off-by: Tim Beermann <beermann@osism.tech>

### DIFF
--- a/roles/zuul/defaults/main.yml
+++ b/roles/zuul/defaults/main.yml
@@ -63,19 +63,20 @@ tenants:
 
 docker_registry: index.docker.io
 
-nodepool_tag: "6.0.0"
-zookeeper_tag: "3.7.0"
+zuul_nodepool_tag: "6.0.0"
+zuul_zookeeper_tag: "3.7.0"
 zuul_tag: "6.0.0"
+zuul_mariadb_tag: "10.6"
+zuul_log_server_tag: "alpine"
 
-image_list:
-  nodepool_builder: "{{ docker_registry }}/zuul/nodepool-builder:{{ nodepool_tag }}"
-  nodepool_launcher: "{{ docker_registry }}/zuul/nodepool-launcher:{{ nodepool_tag }}"
-  zuul_executor: "{{ docker_registry }}/zuul/zuul-executor:{{ zuul_tag }}"
-  zuul_web: "{{ docker_registry }}/zuul/zuul-web:{{ zuul_tag }}"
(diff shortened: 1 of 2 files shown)


roles/zuul/defaults/main.yml
```diff
@@ -67,15 +67,14 @@ nodepool_tag: "6.0.0"
...
-  mariadb: "{{ docker_registry }}/mariadb:latest"
-  log_server: "{{ docker_registry }}/httpd:alpine"
+nodepool_builder_image: "{{ docker_registry }}/zuul/nodepool-builder:{{ nodepool_tag }}"
+nodepool_launcher_image: "{{ docker_registry }}/zuul/nodepool-launcher:{{ nodepool_tag }}"
+zuul_executor_image: "{{ docker_registry }}/zuul/zuul-executor:{{ zuul_tag }}"
+zuul_web_image: "{{ docker_registry }}/zuul/zuul-web:{{ zuul_tag }}"
+zuul_scheduler_image: "{{ docker_registry }}/zuul/zuul-scheduler:{{ zuul_tag }}"
+zookeeper_image: "{{ docker_registry }}/zookeeper:{{ zookeeper_tag }}"
+mariadb_image: "{{ docker_registry }}/mariadb:latest"
+log_server_image: "{{ docker_registry }}/httpd:alpine"
```
osfrickler, line 77: For the last three names maybe we should prefix them with `zuul_` in order to make them unique to this role? Same for `zookeeper_tag` and possibly for the `nodepool*` vars?

While we don't currently deploy this role together with anything else, it could help to avoid future conflicts.

And while we're touching this, maybe also pin mariadb to a specific tag?

roles/zuul/templates/docker-compose.yaml.j2
```diff
@@ -2,11 +2,11 @@ version: '3'
 services:
   mariadb:
     container_name: "{{ container_name.mariadb }}"
-    image: {{ image_list.mariadb }}
+    image: "{{ mariadb_image }}"
```
osfrickler, line 5: Why are you adding all these quotes? Seems to be working fine without them.

tibeer: It does. But all other repos use quotes, too. I just tried to keep it clean ^^

roles/zuul/defaults/main.yml
tibeer: Good idea. Will do this.

PR #868, opened 2022-11-08
Let python requests use the CA bundle

reqa:
Signed-off-by: Arvid Requate <requate@univention.de>

### DIFF
--- a/roles/openstackclient/templates/docker-compose.yml.j2
+++ b/roles/openstackclient/templates/docker-compose.yml.j2
@@ -7,6 +7,8 @@ services:
     restart: unless-stopped
     image: "{{ openstackclient_image }}"
     command: sleep infinity
+    env_file:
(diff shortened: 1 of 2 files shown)


roles/openstackclient/templates/docker-compose.yml.j2
```diff
@@ -7,6 +7,8 @@ services:
     restart: unless-stopped
     image: "{{ openstackclient_image }}"
     command: sleep infinity
+    env_file:
+      - "{{ openstackclient_configuration_directory }}/openstackclient.env"
```
berendt, line 11: Use environment instead of env_file or add a task to copy the openstackclient.env.j2 file. Would prefer to use ``environment``.

berendt (changes requested):

berendt:
Replaced by https://github.com/osism/ansible-collection-services/pull/918.

PR #886, opened 2022-12-02
openstack-health-monitor: use docker-compose service

berendt:

### DIFF
--- a/roles/openstack_health_monitor/README.rst
+++ b/roles/openstack_health_monitor/README.rst
@@ -114,7 +114,12 @@ The container image to use.
 .. zuul:rolevar:: openstack_health_monitor_container_name
    :default: openstack_health_monitor
 
-Name of the container in which Openstack-health-monitor will run.
+Name of the container.
+
+.. zuul:rolevar:: openstack_health_monitor_service_name
+   :default: docker-compose@openstack_health_monitor
+
+Name of the service.
 
 .. zuul:rolevar:: openstack_health_monitor_network
    :default: 172.31.100.160/28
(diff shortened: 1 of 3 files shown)


roles/openstack_health_monitor/README.rst
```diff
@@ -114,9 +114,14 @@ The container image to use.
...
+   :default: openstack_health_monitor
+
+Name of the service.
 
 .. zuul:rolevar:: openstack_health_monitor_network
    :default: 172.31.100.160/28
 
-The network to use for the Openstack-health-monitor container.
\ No newline at end of file
+Name of the network.
```
osfrickler, line 127: An address is not a name IMO. Also what is "the network"? Why not just leave the old description in place?

roles/openstack_health_monitor/defaults/main.yml
```diff
@@ -55,3 +55,5 @@ openstack_health_monitor_image: "{{ docker_registry_openstack_health_monitor }}/
 openstack_health_monitor_container_name: openstack_health_monitor
 
 openstack_health_monitor_network: 172.31.100.160/28
+
+openstack_health_monitor_service_name: "docker-compose@openstack_health_monitor"
```
osfrickler, line 59: Not sure whether an extra variable for this is useful, the@xxx part at least always needs to match the docker_compose_directory above.

berendt: We use a dedicated variable everywhere else for this. We can cleanup this in a separate PR.

berendt: https://github.com/osism/issues/issues/375

osfrickler (approved):

PR #907, opened 2022-12-20
Revert molecule version

ra-beer:
Since all Tests run into errors and there are bug reports
in the molecuke repo, we decide in Slack that we revert to
the older version of molecule.


berendt:
Is this still necessary after we have changed Molecule to delegated?

PR #927, opened 2023-01-24
Refactor wireguard role to allow multiple clients

osfrickler:
In a production environment, we want to allow for multiple users to be able to connect to the wireguard VPN simultaneously, so we add the option to create a peer configuration for each of them.

Also adapt the role to running on recent Ubuntu versions, where the module is available in the kernel by default.


osfrickler:
Testbed adaption in https://github.com/osism/testbed/pull/1525 needs to be merged together with this in order to avoid disruption.

PR #928, opened 2023-01-25
dnsdist: allow multiple host bindings with dnsdist_hosts

berendt:

### DIFF
--- a/roles/dnsdist/README.rst
+++ b/roles/dnsdist/README.rst
@@ -54,6 +54,11 @@ Name from the container in which Dnsdist will run.
 
 The host where Dnsdist will be reachable.
 
+.. zuul:rolevar:: dnsdist_hosts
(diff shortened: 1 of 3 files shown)


roles/dnsdist/README.rst
```diff
@@ -54,6 +54,11 @@ Name from the container in which Dnsdist will run.
 
 The host where Dnsdist will be reachable.
 
+.. zuul:rolevar:: dnsdist_hosts
+   :default: - "{{ dnsdist_host }}"
```
osfrickler, line 58: Shouldn't we at least deprecate dnsdist_host then?

berendt: Does not matter. If you have a single host, use dnsdist_host. Otherwise, overwrite everything with dnsdist_hosts. This way we are backwards compatible and don't have to change anything.

PR #931, opened 2023-01-30
manager: check if /usr/local/bin/osism exsists

berendt:
Closes osism/issues#418


berendt:
This will not resolve the issue. The container is missing, not the wrapper script.

PR #956, opened 2023-02-17
Yamllint: Refactor check

ra-beer:
- Delete check via github action
- Enable yamllint via zuul
- Partly osism/issues#396


berendt:
Makes no sense to me. Why have we introduced the yamllint job in osism/zuul-jobs when we do not use the job?

I think https://github.com/osism/ansible-collection-commons/pull/443 is the better way to use yamllint. This way we have it the same way like for ansible-lint.

PR #959, opened 2023-02-20
Add Zuul installation guide

osfrickler:
Since the successful installation of a Zuul deployment is a bit more complicated than simply applying the service role, we document it in a dedicated chapter of the documentation.


ra-beer:
For publishing in docusaurus and within the scs documentation, we need to rebuild the documentation. Files must be markdown and structure must fitting to this guide: https://github.com/SovereignCloudStack/docs/blob/main/community/contribute/adding-docs-guide.md

osfrickler:
Can the rebuild be done in a second step? Currently everything is still in rst afaict.

berendt:
I think it's fine to keep RST/Sphinx here for the moment. This is the reference documentation of the roles which is independent of the SCS documentation.

berendt:
We will test it with Benjamin with the SCS Zuul and in the TSI project.

PR #960, opened 2023-02-21
Enable proxy for container pulling

tibeer:
In order to force docker to pull only via a proxy, the service unit needs to be altered.

Signed-off-by: Tim Beermann <beermann@osism.tech>


tibeer:
Do not merge! This would break the testbed currently, as the mirrors are not prepared yet.

roles/squid/defaults/main.yml
```diff
@@ -29,14 +32,16 @@ squid_image: "{{ docker_registry_squid }}/ubuntu/squid:{{ squid_tag }}"
...
-  - minio.services.osism.tech
-  - archive.ubuntu.com
-  - download.docker.com
-  - packagecloud.io
-  - github.com
-  - galaxy.ansible.com
-  - pypi.org
-  - .opendev.org
+#  - minio.services.osism.tech
+#  - archive.ubuntu.com
```
osfrickler, line 41: This change is only for containers, I think we'll need to keep the other targets for a bit longer?

tibeer:
recheck

tibeer:
converted to draft again. daemon uses proxy, but clients that spawn containers not. they should pass the http_proxy stuff to every container on start

tibeer:
@berendt this is ready to be merged.

PR #1025, opened 2023-04-20
replace: haveged with rng tools

matfechner:
for a enhanced  entropy behavior for workload instances rng-tools must be enabled on nodes

### DIFF
--- a/roles/rng/README.rst
+++ b/roles/rng/README.rst
@@ -1,14 +1,12 @@
-This ansible role install Haveged.
-Haveged is a random number generator.
+This ansible role install rng-tools.
+rngd is a random number generator.
+
+rng-tools include the Daemon to use a Hardware TRNG random number generator.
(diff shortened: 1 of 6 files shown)


berendt (changes requested):
You also have to uninstall havaged.

roles/rng/README.rst
```diff
@@ -1,14 +1,17 @@
...
+This ansible role install rng.
+rngd is a random number generator.
+
+rng-tools include the Daemon to use a Hardware TRNG random number generator.
+The rngd daemon acts as a bridge between a Hardware TRNG.
 
 **Role Variables**
 
 .. zuul:rolevar:: rng_service_name
    :default: haveged
```
berendt, line 10: This should be rngd. While touching this I think it is fine to remove the rng_servie_name parameter and to hardcode rngd inside the task.

berendt (approved):

PR #1031, opened 2023-04-27
squid: update allowed addresses

osfrickler:
Both quay.io and archive.ubuntu.com have subdomains that may be used in downloading content, like cdn03.quay.io and de.archive.ubuntu.com. Allow all subdomains to avoid blocking things.


artificial-intelligence:
LGTM

artificial-intelligence (approved):

PR #1058, opened 2023-05-26
frr: Add backup config for loadbalancer

osfrickler:
Allow loadbalancers to advertise IPs with a higher metric in order to support active/passive scenarios.

### DIFF
--- a/roles/frr/templates/frr_loadbalancer.conf.j2
+++ b/roles/frr/templates/frr_loadbalancer.conf.j2
@@ -36,23 +36,41 @@ router bgp {{ frr_local_as }}
 exit
 !
 ip prefix-list internal seq 5 permit {{ frr_internal_v4 }} ge 32 le 32
+{% for item in frr_internalbackup_v4 %}
+ip prefix-list internalbackup seq {{ item.seq }} permit {{ item.cidr }} ge 32 le 32
+{% endfor %}
 ip prefix-list external seq 5 permit {{ frr_external_v4 }} ge 32 le 32
 {% for item in frr_externalbackup_v4 %}
 ip prefix-list externalbackup seq {{ item.seq }} permit {{ item.cidr }} ge 32 le 32
 {% endfor %}
 !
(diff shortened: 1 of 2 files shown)


roles/frr/templates/frr_loadbalancer.conf.j2
```diff
@@ -36,27 +36,63 @@ router bgp {{ frr_local_as }}
 exit
 !
 ip prefix-list internal seq 5 permit {{ frr_internal_v4 }} ge 32 le 32
+{% for item in frr_internalbackup_v4 %}
```
berendt, line 39: frr_internalbackup_v4 has to be added to defaults.yml

roles/frr/templates/frr_loadbalancer.conf.j2
```diff
@@ -36,27 +36,63 @@ router bgp {{ frr_local_as }}
...
+ip prefix-list internalbackup seq {{ item.seq }} permit {{ item.cidr }} ge 32 le 32
+{% endfor %}
 ip prefix-list external seq 5 permit {{ frr_external_v4 }} ge 32 le 32
+{% for item in frr_externalbackup_v4 %}
```
berendt, line 43: frr_externalbackup_v4 has to be added to defaults.yml

berendt (changes requested):

roles/frr/templates/frr_loadbalancer.conf.j2
osfrickler: I think it doesn't, because the default service type is leaf

roles/frr/templates/frr_loadbalancer.conf.j2
osfrickler: dito

roles/frr/templates/frr_loadbalancer.conf.j2
berendt: I think it is best practice if used variables are initialized with a value that the template is usable.

berendt (approved):

PR #1064, opened 2023-06-01
manager: avoid 2nd manager restart

berendt:


berendt:
> This looks like a hack to me, but I don't have a better idea out of the box

Yes. But works :)

PR #1073, opened 2023-06-08
chrony: fix use of local chrony template file

berendt:


osfrickler (approved):
Meh, I've been looking at this block multiple times and didn't spot the mismatch, thx for this

PR #1078, opened 2023-06-16
Ansible-lint: Refactor rules and noqas

ra-beer:
- Fix some lintingerrors after refactoring.
- Exclude only affected tasks where necessary.
- Delete not needed noqas
- Partly osism/issues#496

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,8 +1,6 @@
 ---
 exclude_paths:
   - .github/
-  - roles/docker/tasks/install-docker-Debian.yml  # We must exclude this File until we have something that can handle blocks.
-  - roles/nexus/tasks/initialize.yml  # We must exclude this File until we have something that can handle blocks.
 mock_modules:
   - ansible.posix.mount
   - ansible.posix.synchronize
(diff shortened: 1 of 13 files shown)


ra-beer:
Broken pipeline depends on https://github.com/osism/ansible-collection-commons/pull/497

osfrickler:
recheck

.ansible-lint
```diff
@@ -20,11 +18,8 @@ rulesdir:
   - ./.ansible-lint-rules/
 skip_list:
   - galaxy[no-changelog]
-  - ignore-errors
-  - no-handler
-  - package-latest
```
osfrickler, line 25: I would prefer to keep this rule in the skip_list, as we have no intention of following it ever.

osfrickler (commented):
ansible-lint job is still failing

.ansible-lint
ra-beer: The rule that you want to skip is our own rule, which we had include because we want to use all fqcn in all repos. I actually work on this issue.

ra-beer: Done

osfrickler (approved):

PR #1092, opened 2023-06-28
Add tuned role

berendt:


roles/tuned/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+  ansible.builtin.shell: |
+    set -o pipefail
+    tuned-adm active | awk -F': ' '{print $2}'
+  register: result
+  args:
+    executable: /bin/bash
+
+- name: Set tuned profile
+  become: true
+  ansible.builtin.command: tuned-adm profile virtual-host
```
osfrickler, line 23: use tuned_profile here instead of fixed value

roles/tuned/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+  when: tuned_profile != result.stdout
```
osfrickler, line 24: add "changed_when: true"

roles/tuned/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+    enabled: true
+
+- name: Get active tuned profile
+  become: true
+  ansible.builtin.shell: |
+    set -o pipefail
+    tuned-adm active | awk -F': ' '{print $2}'
+  register: result
+  args:
+    executable: /bin/bash
```
osfrickler, line 19: add "changed_when: false"

osfrickler (changes requested):

PR #1094, opened 2023-07-01
squid: whitelist regio.digital services

berendt:


roles/squid/defaults/main.yml
```diff
@@ -40,7 +40,9 @@ squid_allowed_addresses:
   - harbor.services.osism.tech
   - index.docker.io
   - minio.services.osism.tech
+  - osism.harbor.regio.digital
```
berendt, line 43: We should use registry.regio.digital here.

osfrickler (approved):

PR #1095, opened 2023-07-02
squid: add volume to store the squid cache

berendt:


roles/squid/templates/osism.conf.j2
```diff
@@ -0,0 +1,6 @@
+cache_mem 128 MB
+minimum_object_size 0 bytes
+maximum_object_size 500 MB
+maximum_object_size_in_memory 512 KB
```
artificial-intelligence, line 4: maybe make this bigger? but I don't know about the machine sizing or if this is a default or which kind of files will be served through this, so feel free to ignore this.

artificial-intelligence (approved):
LGTM

roles/squid/templates/osism.conf.j2
berendt: APT packages only. I think 500 MByte and 512 KByte in memory is enough for this use case.

PR #1126, opened 2023-08-08
Letsencrypt for standalone zuul

master-caster:
* update compose template to add certificates to zuul_log_server
    container
* update apache configuration to accept acme challenge

Use this as a general discussion whether this is useful in this place or not.
For SCS zuul was deployed using zuul role provided by OSISMs ansible-collection-services.
Since we didn't use all of the services we just used the zuul role that
doesn't provide letsencrypt support.

This patch is not perfect and doesn't solve the chicken-and-egg situation for initial deployment
when certbot couldn't obtain a valid certificate since the zuul_log_server container isn't running and won't run
until certificates are provided.

So many tasks left to do on this if we want to make it right.
But before we invest a lot of effort in here we should discuss if it is the right place. Maybe
we should use another proxy service in front of the zuul_log_server which itself is already a proxy?

Maybe we should switch from http-01 challenge to DNS challenge?

### DIFF
--- a/roles/zuul/defaults/main.yml
+++ b/roles/zuul/defaults/main.yml
@@ -12,6 +12,7 @@ zuul_logserver_fqdn: "logs.example.org"
 zuul_webserver_fqdn: "zuul.example.org"
 zuul_webserver_admin: "admin@example.org"
 zuul_zookeeper_fqdn: "zookeeper01.example.org"
+zuul_use_acme: false
 
 zuul_base_conf_dir: /opt/zuul
 zuul_component_conf_dirs:
--- a/roles/zuul/templates/docker-compose.yaml.j2
+++ b/roles/zuul/templates/docker-compose.yaml.j2
@@ -87,8 +87,7 @@ services:
     volumes:
       - "{{ zuul_component_conf_dirs.log_server }}:/usr/local/apache2/conf:ro"
       - "log_server:/usr/local/apache2/htdocs:z"
-      - "/etc/letsencrypt/live/{{ zuul_logserver_fqdn }}/fullchain.pem:/etc/letsencrypt/live/{{ zuul_logserver_fqdn }}/fullchain.pem:ro"
-      - "/etc/letsencrypt/live/{{ zuul_logserver_fqdn }}/privkey.pem:/etc/letsencrypt/live/{{ zuul_logserver_fqdn }}/privkey.pem:ro"
+      - "/etc/letsencrypt:/etc/letsencrypt:ro"
       - "/opt/zuul/log_server/.well-known/acme-challenge/:/usr/local/apache2/htdocs/.well-known/acme-challenge/:rw"
 
 volumes:
--- a/roles/zuul/templates/docker-compose.yaml.j2
+++ b/roles/zuul/templates/docker-compose.yaml.j2
@@ -3,14 +3,14 @@ services:
   mariadb:
     container_name: "{{ container_name.mariadb }}"
     image: "{{ zuul_mariadb_image }}"
-    env_file: "{{ zuul_base_conf_dir }}/mariadb.env"
+    env_file: "{{ base_conf_dir }}/mariadb.env"
   zookeeper:
     container_name: "{{ container_name.zookeeper }}"
     image: "{{ zuul_zookeeper_image }}"
-    hostname: {{ zuul_zookeeper_fqdn }}
+    hostname: {{ zookeeper_fqdn }}
     volumes:
-      - "{{ zuul_component_conf_dirs.certs }}:/var/certs:z"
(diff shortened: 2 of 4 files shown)


berendt:
Please rebase on main. It includes a lot of already merged commits.

master-caster:
should be up to date now

berendt:
There are still 7 commits in this PR. 4 by myself.

master-caster:
hopefully it is clean now :sweat_smile:

osfrickler:
> hopefully it is clean now sweat_smile

There are still two commits instead of one, the commit messages aren't really helpful and using your internal email address for the signed-off-by seems weird, but let's discuss more important things first.

> But before we invest a lot of effort in here we should discuss if it is the right place. 
> Maybe we should use another proxy service in front of the zuul_log_server which itself is 
> already a proxy?
> 
> Maybe we should switch from http-01 challenge to DNS challenge?

I would in fact prefer to see an implementation automating dns-01 auth, but as it seems that is not a feasible approach for all installation. So adding the option to use http-01 instead is fine for me, but we should make sure that it stays optional.

The webserver container being deployed is already running as a proxy for the zuul backend, adding yet another proxy in front of it doesn't seem helpful to me, and with this patch it also doesn't seem necessary.

Did you already perform some manual testing with certbot and this patch to see that it will work in the end? At least I would not see any obstacles to pursuing the proposed approach.

master-caster:
I've tested the change by removing the running container and recreate it with the ansible role.
Compose file looks good. 
The only issue right now is a chicken-egg-problem. The container won't start if the certificate files are not available.
That could be mitigated by providing a snake oil cert.

master-caster:
For now i didn't find a good and clean solution for the chicken-egg problem regarding the le-certifactes.
I proposed changes to the scs zuul repository to mitigate this problem that, at least i think so, are not
qualified for osism upstream.

You may want to rely on the original behaviour instead of being forced to provide lets encrypt certificates.
@osfrickler may you re check?

roles/zuul/templates/docker-compose.yaml.j2
```diff
@@ -87,6 +87,8 @@ services:
     volumes:
       - "{{ zuul_component_conf_dirs.log_server }}:/usr/local/apache2/conf:ro"
       - "log_server:/usr/local/apache2/htdocs:z"
+      - "/etc/letsencrypt:/etc/letsencrypt:ro"
+      - "/opt/zuul/log_server/.well-known/acme-challenge/:/usr/local/apache2/htdocs/.well-known/acme-challenge/:rw"
```
osfrickler, line 91: I wonder if this should also be optional depending on `zuul_use_acme`. There also should be tasks that ensure that these directories exist on the host.

roles/zuul/templates/httpd.conf.j2
```diff
@@ -380,7 +380,10 @@ SSLStaplingCache "shmcb:/usr/local/apache2/logs/ssl_stapling(32768)"
 ##
 
 <VirtualHost _default_:80>
+Alias /.well-known/acme-challenge/ /usr/local/apache2/htdocs/.well-known/acme-challenge/
```
osfrickler, line 383: Maybe instead just set the DocumentRoot like for the 443 vhost?

roles/zuul/templates/httpd.conf.j2
```diff
@@ -380,7 +380,10 @@ SSLStaplingCache "shmcb:/usr/local/apache2/logs/ssl_stapling(32768)"
...
 RewriteEngine On
+RewriteCond %{REQUEST_URI} /\.well\-known/acme\-challenge/
+RewriteRule (.*) /.well-known/acme-challenge/$1 [L,QSA]
```
osfrickler, line 386: Make this conditional on `zuul_use_acme` too

roles/zuul/defaults/main.yml
```diff
@@ -12,6 +12,7 @@ zuul_logserver_fqdn: "logs.example.org"
 zuul_webserver_fqdn: "zuul.example.org"
 zuul_webserver_admin: "admin@example.org"
 zuul_zookeeper_fqdn: "zookeeper01.example.org"
+zuul_use_acme: false
```
osfrickler, line 15: This var should be mentioned in the README.rst

Ideally an example how to set this up in doc/source/zuul would also be nice, but that can be done as follow-up.

berendt:
I would like to speed things up here and unblock. After internal discussion, we do not see Letsencrypt in this role because we do not use it in this way. I think it's better if we create a fork of this role to implement SCS specific things there. HA specific things or multi-node installations are also not in the scope of this role as we explicitly run single node deployments with this role.

berendt:
Close this here now as we will not be taking it any further in our role at the moment.

PR #1136, opened 2023-08-30
rsyslog: make additional log server protocol configurable

berendt:
Closes osism/issues#646

### DIFF
--- a/roles/rsyslog/defaults/main.yml
+++ b/roles/rsyslog/defaults/main.yml
@@ -14,8 +14,15 @@ rsyslog_package_name: rsyslog
 
 rsyslog_fluentd: true
 
-rsyslog_fluentd_port: 5140
 rsyslog_fluentd_host: 127.0.0.1
+rsyslog_fluentd_port: 5140
 
 fluentd_port: "{{ rsyslog_fluentd_port }}"
 fluentd_host: "{{ rsyslog_fluentd_host }}"
+
+##########################
+# additional rsyslog
+
+rsyslog_additional_protocol: udp
+rsyslog_additional_host: ""
+rsyslog_additional_port: 5140
--- a/roles/rsyslog/templates/71-additional-log-server.conf.j2
+++ b/roles/rsyslog/templates/71-additional-log-server.conf.j2
@@ -1,1 +1,5 @@
+{% if rsyslog_additional_protocol|default('udp') == 'udp' %}
 *.* @{{ rsyslog_additional_host }}:{{ rsyslog_additional_port }}
+{% else %}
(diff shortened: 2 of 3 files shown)


roles/rsyslog/templates/71-additional-log-server.conf.j2
```diff
@@ -1 +1,5 @@
+{% if rsyslog_additional_protocol|default('udp') == 'udp' %}
```
osfrickler, line 1: Add this var to the role doc?

It is also a bit optimistic to treat everything != udp as tcp, but I'm also not sure it is worth an extra check.

roles/rsyslog/templates/71-additional-log-server.conf.j2
```diff
@@ -1 +1,5 @@
...
 *.* @{{ rsyslog_additional_host }}:{{ rsyslog_additional_port }}
+{% else %}
+*.* @@{{ rsyslog_additional_host }}:{{ rsyslog_additional_port }}
```
artificial-intelligence, line 4: LGTM, but:

according to this [post](https://www.rsyslog.com/sending-messages-to-a-remote-syslog-server/) this is a legacy format that should not be used anymore.
It seems that the new format is more readable. so we might want to adapt this in the future.

artificial-intelligence (approved):

roles/rsyslog/templates/71-additional-log-server.conf.j2
berendt: @artificial-intelligence Please open an issue for this task in osism/issues.

artificial-intelligence: https://github.com/osism/issues/issues/654

roles/rsyslog/defaults/main.yml
```diff
@@ -14,8 +14,15 @@ rsyslog_package_name: rsyslog
...
+rsyslog_fluentd_port: 5140
 
 fluentd_port: "{{ rsyslog_fluentd_port }}"
 fluentd_host: "{{ rsyslog_fluentd_host }}"
+
+##########################
+# additional rsyslog
+
+rsyslog_additional_protocol: udp
+rsyslog_additional_host: ""
```
osfrickler, line 27: We just had a discussion in kolla in a similar situation. I think it is better to leave the variable undefined by default, the code does already check whether it is defined.

berendt: CI is not happy with undefined variable because of the |length check in the template.

osfrickler: hmm, o.k., let's leave it like this for now

osfrickler (approved):

PR #1249, opened 2024-01-02
squid: remove dstdomain whitelisting

berendt:


roles/squid/tasks/main.yml
```diff
@@ -26,6 +26,11 @@
     - osism.conf
```
osfrickler, line 26: need to drop it from this loop, too

osfrickler (commented):
thinking about this, having an ACL does also serve the purpose of protecting from running an open proxy in case the squid process is by accident accessible from a public network. so if we proceed with this, we might need to introduce an ACL based on source IP instead that defaults to 127.0.0.1 or maybe RFC1918 space?

berendt:
> thinking about this, having an ACL does also serve the purpose of protecting from running an open proxy in case the squid process is by accident accessible from a public network. so if we proceed with this, we might need to introduce an ACL based on source IP instead that defaults to 127.0.0.1 or maybe RFC1918 space?

Makes sense to me to add this in the future.

osfrickler (approved):

PR #1254, opened 2024-01-07
wireguard: use no_log for "Copy client configuration files"

berendt:
This way the preshared key will not be logged.


osfrickler (approved):
The client key is not written into this file, it is private to the user, who only need to share their public key for this config:

```
PrivateKey = CHANGEME - {{ item.name }} private key
```

But your argument does hold true for the preshared key, which is indeed included in the client config file. So adding the `no_log` is still warranted, just with a different reasoning. Maybe amend the commit message accordingly?

osfrickler (approved):

PR #1264, opened 2024-01-16
docker: move proxy settings to the daemon.json file

berendt:


roles/docker/templates/daemon.json.j2
```diff
@@ -50,6 +50,13 @@
 
 {% endfor %}
     },
+{% if docker_configure_proxy|bool %}
```
osfrickler, line 53: Add spaces around `|` for readability?

roles/docker/templates/daemon.json.j2
```diff
@@ -50,6 +50,13 @@
...
+    "proxies": {
+      "http-proxy": "{{ docker_proxy_http }}",
+      "https-proxy": "{{ docker_proxy_https }}",
+      "no-proxy": "{{ docker_proxy_no_proxy|join(',') }}"
```
osfrickler, line 57: dito

osfrickler (commented):
wow, the docker config file format is really simple. just use the same keys that you would use as command line options. except for those funny random options that are special and need to go into a subgroup. which ones is affected is only documented in a sample config file

Issue #1265, opened 2024-01-17
openstack CLI docker tty


Nils98Ar:
The `-t` leads to adding a carriage return to the output when running the `openstack` command interactively.
Why would you want to run the openstack commands in the container with tty?

Because of that we need a `tr -d "\r"` before working with the output all the time.

https://github.com/osism/ansible-collection-services/blob/2a9f8a404547e98ad7ee5cc2628018bb59836a25/roles/openstackclient/templates/openstack.j2#L13C12-L13C12

See https://github.com/moby/moby/issues/8513#issuecomment-216191236.

berendt:
At the moment we use `-it` everywhere when we run a command inside a container. There is no specific reason, it has worked so far. Feel free to change it where required.

osfrickler:
Removing the `-it` globally breaks interactive clients and also terminal width detection with the `--fit-width` option for the openstack client. You can either add `</dev/null` to the client invocation or we can add a fix like proposed in https://github.com/osism/ansible-collection-services/pull/1268.

PR #1266, opened 2024-01-17
Remove -it for docker exec command

hosfeld:
docker exec -it adds \r\n newlines on each line, making shell parsing more difficult. 

See https://github.com/osism/ansible-collection-services/issues/1265


berendt:
Remove it everywhere + add a release note with `reno new remove-docker-it-arguments` or something like this.

roles/openstackclient/templates/openstack.j2
```diff
@@ -2,17 +2,8 @@
 
 # DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
 
-INTERACTIVE=${INTERACTIVE:-false}
 PARAMETERS=${PARAMETERS:-}
 
-if [[ -t 0 ]]; then
```
osfrickler, line 8: This tests whether stdin is a terminal. So if you want to have the output in unix format, just invoke `openstackclient </dev/null`

osfrickler (commented):
I'm not sure whether this is a good idea. It breaks using the client in interactive mode, which may or may not be very useful, but at least it is an option currently.

roles/openstackclient/templates/openstack.j2
hosfeld: We can either leave the current settings and remove the \r\n from output:
`for i in $(openstack server list -c ID -f value | tr -d "\r"); do openstack server show $i; done`

Or we change the default behaviour of the docker-wrapper-scripts to not containing carriage-returns for easier scripting usability (and maybe add an additional flag if you really want to run an interactive session?)

osfrickler: Not sure what is wrong with adding the `</dev/null`, but maybe https://github.com/osism/ansible-collection-services/pull/1268 suits you better?

PR #1268, opened 2024-01-17
docker wrappers: make INTERACTIVE=false work

osfrickler:
Allow the override via environment to work in both directions.


osfrickler:
will adapt all scripts accordingly if we agree to this approach

berendt (approved):

PR #1271, opened 2024-01-24
docker: ensure that some packages are not installed

berendt:
Over time, the dependencies in the Docker repository have changed. It should therefore be ensured that packages from the past no longer remain installed.

### DIFF
--- a/roles/docker/tasks/install-docker-Debian.yml
+++ b/roles/docker/tasks/install-docker-Debian.yml
@@ -180,3 +180,11 @@
   when:
     - docker_registry_username is defined and docker_registry_username|string|length > 0
     - docker_registry_password is defined and docker_registry_password|string|length > 0
+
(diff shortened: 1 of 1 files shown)


roles/docker/tasks/install-docker-Debian.yml
```diff
@@ -180,3 +180,9 @@
   when:
     - docker_registry_username is defined and docker_registry_username|string|length > 0
     - docker_registry_password is defined and docker_registry_password|string|length > 0
+
+- name: Ensure that some packages are not installed
+  become: true
+  ansible.builtin.apt:
+    name: ["docker-ce-rootless-extras", "docker-buildx-plugin"]
```
osfrickler, line 187: nits: wrong order (b<c), also make an explicit list with one entry per line to easy future changes?

osfrickler (approved):

PR #1273, opened 2024-01-25
Fix Ansible-lint issues without rule var-naming[no-role-prefix] on skip_list

sbstnnmnn:
cf. https://github.com/osism/issues/issues/829


.ansible-lint
```diff
@@ -8,7 +8,7 @@ rulesdir:
   - ./.ansible-lint-rules/
 skip_list:
   - galaxy[no-changelog]
-  - var-naming
+# - var-naming
```
sbstnnmnn, line 11: rm commented line

lindenb1 (approved):
Looks good to me

jklare:
Hi @sbstnnmnn @lindenb1 , I am currently going through the open issues in our repositories and found this one. Looking at the merge conflicts here and given that this PR seems to target linting issues, I am wondering if this is still relevant and if you want to continue working on this. If so, this should be rebased before the next review and merge. If not, I am fine if we close this issue for now.

berendt:
That's just leftover stuff and can be thrown away. This approach would have too great an impact on the existing configurations.

PR #1290, opened 2024-02-12
Fixing falco role

lindenb1:
cf. https://github.com/osism/issues/issues/850


sbstnnmnn (commented):
May summarize to one task:

```
- name: Specify driver type with falcoctl and install falco driver
  become: true
  ansible.builtin.command:
    cmd: |
      falcoctl driver config --type kmod
      falcoctl driver install
  notify: "Restart service {{ falco_service_name }}"
  changed_when: true
```

PR #1298, opened 2024-02-14
add missing template for cloudnative-pg

reqa:
For https://github.com/SovereignCloudStack/issues/issues/507


berendt:
I had intentionally removed the dependency on cloudnative_pg for the moment because we are not yet mapping dependencies via Ansible role meta information. At the moment, we do it through the playbooks only.

PR #1310, opened 2024-02-26
manager: ensure the user for the mariadb healthcheck exists

berendt:
Closes osism/issues#863


berendt:
That's handled by the MariaDB upgrade section in the docker-entry point.sh script when MARIADB_AUTO_UPGRADE=1 in the mariadb.env.

PR #1334, opened 2024-03-05
Remove openldap role

berendt:
Part of osism/issues#822


janhorstmann (changes requested):
[`.zuul.yaml` still has a job defined for the `openldap` role](https://github.com/osism/ansible-collection-services/blob/gh822/.zuul.yaml#L300).  
It can probably be removed as well

janhorstmann (approved):
LGTM

PR #1342, opened 2024-03-18
Remove bird role

berendt:
The integration had been removed for some time. The role will now also be removed.

Closes osism/issues#924


gndrmnn:
This change is now also included in #1284 so this MR is not necessary

gndrmnn:
Obsoleted with #1284

PR #1352, opened 2024-04-02
Fix .zuul.yml files regexes

gndrmnn:


jklare:
Hi @gndrmnn , I was just looking at this and saw that this has been in draft status for about 2 years now. Do you think this should just be updated/rebased and would then be ready to be merged, or should we rather close this in favour of creating a new PR once we find time to work on this again?

gndrmnn:
@jklare The problem this PR was trying to solve is that the existing regex would trigger a CI test for changes to service python files in the `/test` subfolder, but not for other files such as variables in the `/vars` folder. This means that you could create a PR which changes a variable to an invalid value but the CI for that service would not run and you would not see that the service was broken until it was rolled out.

I don't know whether or not this is still relevant, so feel free to close

jklare:
Closing this based on the feedback. If the issue persists, we can pick it up again in a new PR.

PR #1358, opened 2024-04-05
Create FRR template loadbalancer_external_uplink

fzakfeld:
fixes https://github.com/osism/issues/issues/949

We have a setup with dedicated links for external connectivity. The public IP addresses for the loadbalancer are assigned to a separate dummy interface and announced via the dedicated links. The default route for the external connectivity is installed in to the routing table defined with `frr_incoming_table_external`, in order to allow PBR.


roles/frr/templates/frr_loadbalancer_external_uplink.conf.j2
```diff
@@ -0,0 +1,63 @@
...
+route-map bgp_out_internal permit 10
+ match interface {{ frr_dummy_interface_internal }}
+exit
+!
+route-map bgp_out_external permit 10
+ match interface {{ frr_dummy_interface_external }}
+exit
+!
+route-map bgp_in_external permit 10
+ set table {{ frr_incoming_table_external }}
```
berendt, line 50: Add `frr_incoming_table_external` with a sane value to `defaults/main.yml`.

fzakfeld: Now defaulting to 254 (default according to https://man7.org/linux/man-pages/man8/ip-route.8.html#:~:text=By%20default%20all%20normal%20routes,local%20table%20(ID%20255). )

roles/frr/defaults/main.yml
```diff
@@ -29,3 +29,5 @@ frr_externalbackup_v4: []
 frr_externalbackup_v6: []
 frr_internalbackup_v4: []
 frr_internalbackup_v6: []
+
+frr_incoming_table_external: 254
```
berendt, line 33: Ansilbe Linter is unhappy because of the missing \n\r here.

PR #1384, opened 2024-04-27
zuul: install docker with pip for the manager role

berendt:
Since we are working within a venv, but the python-3docker package is installed via apt, it must be installed manually in the venv so that the facts of osism.services.docker work.


berendt:
Recheck

PR #1388, opened 2024-04-27
Remove requirements.yml file

berendt:
We manage the requirements via osism/release. Using main here can lead to strange problems in Zuul.


berendt:
Recheck

PR #1418, opened 2024-05-01
manager: support scripts

berendt:
These scripts combine a few things. They are based on the scripts in the osism/testbed repository.


berendt:
Recheck

PR #1480, opened 2024-06-12
[DRAFT] Use configmap to load various certs

JuanPTM:
Load all CA's from a configmap into keycloak truststore, enabling SSL from keycloak to other services.


roles/keycloak/templates/configmap-certs.yml
```diff
@@ -0,0 +1,69 @@
...
+      3FFzLE5G8tSJGjmRYajQDTCJZPoVmdj/hmFdqGDyPngDWOAZYFNDkU9ES9/zsP++
+      Ayiix/lTlIpLClUcarqW00xfZPk5ieCUNzFJd8K5BPL7nIlj+fFHy0GsIvMirWV7
+      NCKVoUuESpt8hlnRO+gcpKNlSHiKZa/8xsvGp4l2tzH0p3dBiJo0sIdA09D1dd9X
+      jQkHRFBqRQHiXtw7EEIxJE8bW8A7VG+zNVyuat70GWjLof5hruhR5/NGiPzrETWY
+      aKOUxyH2iO99Fmo1aWyp5ILUxB6Q5FH+6VRvONCTcyitc5+K9bBgKnuwcfO5Lfsu
+      mzq17dh1vEfuZB7NpadXsPy9C/Trlymcbj4SFza/z13/kH0rphyrmcBy2vU0/Ht0
+      R8lxKW9RYZSPCM7c3M+dEzKnynxJcKkevpwuyxM3jqWLxxBjiyM932mV/TbXZUhr
+      NckvU3kvKYJLBcZ0sbYVawp8bPV8AH4CrrchH5WD81pNFoMZsL510ua0
+      -----END CERTIFICATE-----
+...
```
berendt, line 69: This looks wrong.

roles/keycloak/templates/configmap-certs.yml
```diff
@@ -0,0 +1,69 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: configmap-certs
+  namespace: keycloak
+
+data:
+  testbed-cert.pem: |
```
berendt, line 9: Do not add testbed files here as hardcoded default. The value of the data dictionary has to be configurable via a parameter. This parameter has then to be added to the osism/testbed repository.

berendt (changes requested):

roles/keycloak/templates/configmap-certs.yml
JuanPTM: File removed

roles/keycloak/templates/configmap-certs.yml
JuanPTM: File completely removed. Opted for a mounted volume with all certs into the initContainer

roles/keycloak/tasks/deploy.yml
```diff
@@ -1,4 +1,10 @@
 ---
+- name: "Create configMap with TLS certificates"
+  kubernetes.core.k8s:
+    state: present
+    kubeconfig: /share/kubeconfig
+    definition: "{{ lookup('file', keycloak_configuration_directory + '/configmap-certs.yml') | from_yaml_all }}"
```
gtema, line 6: I am actually wondering, what is the configmap-certs.yml? It is not part of the PR, not in the repo, not present in the documentation

roles/keycloak/templates/codecentric-keycloak.yml.j2
```diff
@@ -75,6 +75,10 @@ extraVolumes: |
         secretName: {{ '{{ include "keycloak.fullname" . }}' }}-tls-certs
     - name: updated-cacerts
       emptyDir: {}  # work dir for update-truststore.sh
+    - name: certs-dir
```
gtema, line 78: I am wondering about 2 issues:
1) Usually in K8 certs are always handled as secrets and not as cm. I understand that there is no secret in the CA certs, but there is some common practice established
2) `hostPath` is looking very weird to me. If you want to mount configMap you do it like https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#populate-a-volume-with-data-stored-in-a-configmap For secrets it is a bit different but is shown few lines above. Hostpath is used when you want to mount physical folder of the K8 worker node into the pod. It is bad in that sense, that the pod cannot be rescheduled to another node since the data is not there.

gtema (commented):
Sadly reviewing ansible using templates to generate k8 templates used by helm defined somewhere else without tests is absolutely insane. But I have a feeling it is not done optimally.

roles/keycloak/tasks/deploy.yml
JuanPTM: This is a remnant of the first iteration, this was discarded but I didnt clean up all it seems

roles/keycloak/templates/codecentric-keycloak.yml.j2
JuanPTM: 1. Yes, certs are not mounted as configmaps anymore. 
2. That's what I wanted, as the ansible playbook is executed from inside a container, it mounts a physical folder from the manager node on the initContainer only, that containers loads the certs into the truststore and the volume is no longer needed.

gtema: ehh, this is pretty problematic. From the Ansible<=>K8 point of view you should never make an assumption that you execute playbook on the same host as k8 (hostPath may be on a different host) since it may be easily changed and all of a sudden you get absolutely insane bugs not understanding why. I do not think saving one CM/Secret is worth of introducing bugs potential. It is just not what you do typically when provisioning applications to K8.
Imagine another use case where you deploy your applications using ArgoCD or similar stuff - it just runs somewhere and so you setup should be always referring to CMs/Secrets and eventually PVCs populated during init.

JuanPTM: Then WHAT needs to be implemented here? The volume approach might cause problems, the CM approach where the operator has to write the CA's wont work either. So... any ideas?

gtema: Why does the CM approach not work? Maybe I missed something in the discussion. In my eyes the most correct approach is that inventory with which Ansible is triggered should have variable populated with the content of the certs (or it points to the local dir, but then the playbooks and not the role is responsible in reading this in) and those are used to create K8 Secret (or CM, but Secret is more logical). This Secret/CM is mounted as FS into the init container and it reads it in.

gtema: >     2. That's what I wanted, as the ansible playbook is executed from inside a container, it mounts a physical folder from the manager node on the initContainer only, that containers loads the certs into the truststore and the volume is no longer needed.

I guess we have a misunderstanding here: Ansible role (what you are changing) is not a playbook like you say. Role is ONLY a sequence of steps (tasks) applied to the target. Playbook in difference is a mapping of tasks to targets (called plays). So inside the role you actually do not do any assumptions on from where it is executed (there are very rare exceptions to that, but that is advanced), you only describe what to do with the target.

Roles are designed to have access to variables passed to it (by the playbook applying this role to the target), own role defaults and role files/templates (which are "transferred" to the target during execution)

JuanPTM: Well, for me it's enough if someone can explain me where should I put my stuff to work, so I don't have to expend a crazy amount of time trying to figure it out looking into 3 different repos. I only want SSL from Keycloak to work. 

I guess that leave us with some jinja2 template for writing all the certificates into a CM and mount that into the initContainer. Let's hope I can figure it out how to do it and where to place it.

JuanPTM: New version, now a configmap is generated from a j2 template and the files from a configurable directory

JuanPTM:
To QA this you can use the testbed deployment. In the first place you need to use this branch of the testbed https://github.com/osism/testbed/pull/2261

After that you'd need to modify the testbed deployment to use the branch from this PR for the `ansible-collection-services` (Maybe contact someone from osism, on how to do it)

Then deploy the testbed and continue deploying the IAM stack.

as explained here: https://docs.scs.community/docs/iam/

PR #1482, opened 2024-06-13
Add option for host networking instead of container networking

fzakfeld:
Some containers liek cephclient and openstackclient are using container networking, causing the outside of the container to be reached via NAT.

We've ran in to an issue where the NAT was done with the source-IP address of our management interface instead of the dummy interface that should be used, causing the packets to reach it's destination with the wrong source-IP.

I think this might even be a bug in Docker's networking, as it honoured the route but dismissed the 'src' flag. Quite frankly we are fine with running the containers in host networking mode, so there should be an option to do so if desired IMO.

### DIFF
--- a/roles/cephclient/templates/docker-compose.yml.j2
+++ b/roles/cephclient/templates/docker-compose.yml.j2
@@ -5,7 +5,7 @@ services:
     restart: unless-stopped
     image: "{{ cephclient_image }}"
     command: sleep infinity
-{% if cephclient_host_networking %}
+{% if cephclient_host_networking | bool %}
     network_mode: host
 {% endif %}
     volumes:
@@ -13,7 +13,7 @@ services:
       - "{{ cephclient_data_directory }}:/data:rw"
       - "/etc/hosts:/etc/hosts:ro"
 
-{% if not cephclient_host_networking %}
+{% if not cephclient_host_networking | bool %}
 networks:
   default:
     driver: bridge
--- a/roles/cephclient/templates/docker-compose.yml.j2
+++ b/roles/cephclient/templates/docker-compose.yml.j2
@@ -5,11 +5,15 @@ services:
     restart: unless-stopped
     image: "{{ cephclient_image }}"
     command: sleep infinity
+{% if cephclient_host_networking %}
+    network_mode: host
+{% endif %}
     volumes:
       - "{{ cephclient_configuration_directory }}:/etc/ceph:rw"
       - "{{ cephclient_data_directory }}:/data:rw"
       - "/etc/hosts:/etc/hosts:ro"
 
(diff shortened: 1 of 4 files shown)


roles/cephclient/templates/docker-compose.yml.j2
```diff
@@ -19,3 +23,4 @@ networks:
       driver: default
       config:
         - subnet:  {{ cephclient_network }}
+{% endif %}
```
berendt, line 26: Missing \n

roles/openstackclient/templates/docker-compose.yml.j2
```diff
@@ -32,3 +35,4 @@ networks:
       driver: default
       config:
         - subnet:  {{ openstackclient_network }}
+{% endif %}
```
berendt, line 38: Missing \n

roles/openstackclient/templates/docker-compose.yml.j2
```diff
@@ -22,7 +25,7 @@ services:
 {% endif %}
 
 
-
+{% if not openstackclient_host_networking %}
```
berendt, line 28: Use openstackclient_host_networking | bool here

roles/openstackclient/templates/docker-compose.yml.j2
```diff
@@ -5,6 +5,9 @@ services:
     restart: unless-stopped
     image: "{{ openstackclient_image }}"
     command: sleep infinity
+{% if openstackclient_host_networking %}
```
berendt, line 8: Use openstackclient_host_networking | bool here

roles/cephclient/templates/docker-compose.yml.j2
```diff
@@ -5,11 +5,15 @@ services:
...
     command: sleep infinity
+{% if cephclient_host_networking %}
+    network_mode: host
+{% endif %}
     volumes:
       - "{{ cephclient_configuration_directory }}:/etc/ceph:rw"
       - "{{ cephclient_data_directory }}:/data:rw"
       - "/etc/hosts:/etc/hosts:ro"
 
+{% if not cephclient_host_networking %}
```
berendt, line 16: Use cephclient_host_networking | bool here

roles/cephclient/templates/docker-compose.yml.j2
```diff
@@ -5,11 +5,15 @@ services:
     restart: unless-stopped
     image: "{{ cephclient_image }}"
     command: sleep infinity
+{% if cephclient_host_networking %}
```
berendt, line 8: Use cephclient_host_networking | bool here

berendt (changes requested):
See comments.

PR #1506, opened 2024-07-09
hddtemp: Add drivetemp module check

sbstnnmnn:
Check if drivetemp module is available before loading on Debian and Ubuntu.

Reverts changes from https://github.com/osism/ansible-collection-services/pull/1467


lindenb1 (approved):
LGTM

PR #1561, opened 2024-08-14
Fix default SCRIPTS_ROOT path in Netbox configuration.py

matofeder:
The default directory /etc/netbox/scripts does not exist in the NetBox container, therefore, the attempt to add the script failed.

This commit sets the existing directory /opt/netbox/netbox/scripts as the default.

For issue replication:
- Add a new data source to the Netbox (navigate to Operations/Data sources) and sync it.
-  Try to add a script from added data source in Netbox (navigate to Customization/Scripts)

### DIFF
--- a/molecule/delegated/tests/netbox.py
+++ b/molecule/delegated/tests/netbox.py
@@ -101,7 +101,7 @@ def test_netbox_config(host):
     assert config_py.user == get_variable(host, "operator_user")
     assert config_py.group == get_variable(host, "operator_group")
     assert (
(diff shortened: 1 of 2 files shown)


berendt:
@matofeder Also change https://github.com/osism/ansible-collection-services/blob/main/molecule/delegated/tests/netbox.py#L103-L106

matofeder:
> @matofeder Also change https://github.com/osism/ansible-collection-services/blob/main/molecule/delegated/tests/netbox.py#L103-L106

8fdb19f16c5954707c37a67067e80820bc71bcbb

PR #1598, opened 2024-09-17
WIP: Use reset_connection in prepare step

gndrmnn:


berendt:
Does not work. Revert the single step thing in ansilbe-molecule zuul job.

PR #1616, opened 2024-09-24
share pids with host

scoopex:
Scaphandre is only capable to fetch detailed information about process execution for processes which have access to the global pid namespace.

This enables the parameter "pid: host" for the docker compose file. (https://docs.docker.com/reference/cli/docker/container/run/#pid)

### DIFF
--- a/roles/scaphandre/defaults/main.yml
+++ b/roles/scaphandre/defaults/main.yml
@@ -26,6 +26,8 @@ scaphandre_service_name: "docker-compose@scaphandre"
 scaphandre_host: 127.0.0.1
 scaphandre_port: 9155
 
+scaphandre_share_pids_with_host: true
+
 scaphandre_exporter: prometheus
 scaphandre_flags_defaults:
   - "--qemu"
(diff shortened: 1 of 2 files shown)


fkr:
Out of curiosity: in which cases is not **not** wanted to let scaphandre have this access? I'm asking, because a toggle is introduced (which suggests there are cases where this is not wanted).

berendt:
> Out of curiosity: in which cases is not **not** wanted to let scaphandre have this access? I'm asking, because a toggle is introduced (which suggests there are cases where this is not wanted).

If you are not interested in the details of the processes. 
For example, we only use the service to know the total consumption of a node. We don't care about individual processes.

scoopex:
> Out of curiosity: in which cases is not **not** wanted to let scaphandre have this access? I'm asking, because a toggle is introduced (which suggests there are cases where this is not wanted).

There are only a few reasons why you wouldn't want that. If the toggle is deactivated, Scaphandre can only collect general data and not process-specific data of the global system. If Scaphandre uses the same PID namespace as the host system, this may not be wanted in certain cases for security reasons. For this reason, I have made it possible to disable this.

fkr:
thanks for the explanations. I just always question toggles, because often toggles are introduced, despite a real use-case :)

PR #1622, opened 2024-09-25
frr: make it possible to not use ipv6 in the leaf template

berendt:


jklare:
This looks like it definitely needs a rebase. Given that the code here has changed a bit over the last 1.5 years, i am wondering if this is still relevant. Maybe @berendt you can check if this should be rebased or can be closed?

osfrickler:
I think if someone really needs this, they can achieve it by deploying a custom template, so closing for now

PR #1647, opened 2024-10-01
hddtemp: Enhance CI tests

gndrmnn:
https://github.com/osism/issues/issues/1132


gndrmnn:
Can not be functionally tested in CI

PR #1698, opened 2024-11-20
Give more control over frr config

scoopex:
- Allow the user to decide to automatically restart frr after configuration changes or not
- Allow the user to override the location of the config files (this is useful with the previous enhancements to dryrun, compare/inspect and test structural changes)

### DIFF
--- a/roles/frr/defaults/main.yml
+++ b/roles/frr/defaults/main.yml
@@ -42,3 +42,9 @@ frr_enable_bfdd: "no"
 
 frr_options_bgpd: "-A 127.0.0.1"
 frr_options_bfdd: "-A 127.0.0.1"
+
+frr_config_file: "/etc/frr/frr.conf"
+frr_vtysh_file: "/etc/frr/vtysh.conf"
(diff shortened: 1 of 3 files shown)


roles/frr/defaults/main.yml
```diff
@@ -42,3 +42,9 @@ frr_enable_bfdd: "no"
 
 frr_options_bgpd: "-A 127.0.0.1"
 frr_options_bfdd: "-A 127.0.0.1"
+
+frr_config_file: "/etc/frr/frr.conf"
+frr_vtysh_file: "/etc/frr/vtysh.conf"
+frr_daemons_file: "/etc/frr/daemons"
+
+frr_automatic_restart: true
```
berendt, line 50: Use frr_allow_service_restart as parameter name.

berendt (changes requested):

roles/frr/defaults/main.yml
scoopex: Renamed it as requested.

PR #1730, opened 2025-01-04
homer: add cgit dashboard

berendt:
Closes osism/issues#1200

### DIFF
--- a/roles/homer/defaults/main.yml
+++ b/roles/homer/defaults/main.yml
@@ -36,6 +36,7 @@ homer_service_name: "docker-compose@homer"
 
 homer_url_ara: "http://{{ ara_server_host | default(ansible_default_ipv4.address) }}:{{ ara_server_port | default(8120) }}"
 homer_url_ceph: "http://{{ kolla_internal_vip_address }}:8140"
+homer_url_cgit: "http://{{ cgit_host | default(ansible_default_ipv4.address) }}:{{ cgit_port | default(8210) }}"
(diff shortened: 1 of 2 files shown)


roles/homer/defaults/main.yml
```diff
@@ -36,6 +36,7 @@ homer_service_name: "docker-compose@homer"
 
 homer_url_ara: "http://{{ ara_server_host | default(ansible_default_ipv4.address) }}:{{ ara_server_port | default(8120) }}"
 homer_url_ceph: "http://{{ kolla_internal_vip_address }}:8140"
+homer_url_cgit: "http://{{ cgi_host | default(ansible_default_ipv4.address) }}:{{ cgit_port | default(8210) }}"
```
osfrickler, line 39: s/cgi_host/cgit_host/

also, the default in roles/cgit/defaults/main.yml is `127.0.0.1`, better be consistent here?

osfrickler (changes requested):

roles/homer/defaults/main.yml
berendt: The default for all the `_host` parameters is `127.0.0.1`. I think it makes more sense to change the default of those parameters to `ansible_default_ipv4.address`? Or do we change everything here to 127.0.0.1?

berendt: Variable name is fixed.

osfrickler (approved):

PR #1737, opened 2025-01-10
wireguard: use empty string as default for wiregurad_dns

berendt:
This will avoid the following issue:

object of type 'NoneType' has no len(). object of type 'NoneType' has no len()

### DIFF
--- a/roles/wireguard/defaults/main.yml
+++ b/roles/wireguard/defaults/main.yml
@@ -7,4 +7,4 @@ wireguard_server_public_address: WIREGUARD_PUBLIC_IP_ADDRESS
 wireguard_endpoint: "{{ wireguard_server_public_address }}:{{ wireguard_listen_port }}"
 wireguard_create_client_config: false
 wireguard_client_allowed_ips: "192.168.16.0/20,172.31.252.0/23,10.43.0.0/16,192.168.112.0/20,192.168.128.0/20"
-wireguard_dns:
+wireguard_dns: ""


janhorstmann (approved):
LGTM, because of the length check, but why not just leave it completely without default, so that the `is defined` filter should work?

berendt:
> LGTM, because of the length check, but why not just leave it completely without default, so that the `is defined` filter should work?

Not sure if it is a good idea to use a variable in the tasks that is not defined in the defaults.

janhorstmann:
> > LGTM, because of the length check, but why not just leave it completely without default, so that the `is defined` filter should work?
> 
> Not sure if it is a good idea to use a variable in the tasks that is not defined in the defaults.

I am really unsure what would be considered best practice in this case, but for completeness sake I just found this:
https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_templating_undef.html
which allows to explicitly set a variable to type `ansibleUndefined`, e.g.:

```
wireguard_dns: "{{ undef() }}"
```

Requires ansible >= 2.12

berendt:
> > > LGTM, because of the length check, but why not just leave it completely without default, so that the `is defined` filter should work?
> > 
> > 
> > Not sure if it is a good idea to use a variable in the tasks that is not defined in the defaults.
> 
> I am really unsure what would be considered best practice in this case, but for completeness sake I just found this: https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_templating_undef.html which allows to explicitly set a variable to type `ansibleUndefined`, e.g.:
> 
> ```
> wireguard_dns: "{{ undef() }}"
> ```
> 
> Requires ansible >= 2.12

Ah cool. I didn't know that. Maybe that makes sense in various places.

PR #1747, opened 2025-02-03
Remove virtualbmc role

berendt:


janhorstmann (changes requested):
The role is also mentioned in `.github/renovate.json`

janhorstmann (approved):

PR #1775, opened 2025-03-14
Remove usage of `osism_netbox_image` in manager role

janhorstmann:
The `osism_netbox_image` is no longer build and replaced with the generic `osism_image`.

Part of https://github.com/osism/issues/issues/1223


janhorstmann:
Depends on https://github.com/osism/python-osism/pull/1214

PR #1776, opened 2025-03-17
Default to non-interactive mode in wrapper scripts

janhorstmann:
Attaching a terminal during docker exec results in an additional carriage return before newlines ([1]). While this is expected behaviour by TTYs, this leads to unexpected results when piping output andi splitting on newlines.
Therefore we default to non-interactive mode if any of stdin, stdout and stderr is not a terminal.

[1]
https://github.com/moby/moby/issues/37366


janhorstmann:
Without this piping output of the openstack client will not be intuitive

```
$ openstack image list -f value -c ID | cat -A
c70a325f-3e3a-402e-92aa-4123b9547f36^M$
                                       d7bfcc46-64d6-42ce-a512-2893561fa8b3^M$
                                                                              46531d16-a9cf-46b7-b221-534ec93657c0^M$
```

E.g. doing further processing using image IDs will fail

```
$ openstack image list -f value -c ID | while read -r ID; do ...; done
```

roles/openstackclient/templates/openstack.j2
```diff
@@ -4,7 +4,7 @@
 
 PARAMETERS=${PARAMETERS:-}
 
-if [[ -t 0 ]]; then
+if [[ -t 0 && -t 1 && -t 2 ]]; then
```
osfrickler, line 7: I'm not sure if the check for stderr is even needed, this might lead to issues in the opposite direction?

osfrickler (commented):
Seems like a useful idea, I'm just not sure how to properly test it.

Also the change should likely be done consistently for all our wrappers, cf. https://github.com/osism/ansible-collection-services/pull/1268/files

Finally please shorten the commit message header to ideally not more than 50 chars, you can be more verbose in the body

janhorstmann:
> Seems like a useful idea, I'm just not sure how to properly test it.
> 

Maybe I am missing your point, but  shouldn't piping the output into something that shows non-printable charcaters do the trick?

> Also the change should likely be done consistently for all our wrappers, cf. https://github.com/osism/ansible-collection-services/pull/1268/files
> 

I did not apply this to all the deprecated wrappers, but otherwise hope to have got them all

> Finally please shorten the commit message header to ideally not more than 50 chars, you can be more verbose in the body

Done. Sorry about that, when I initially pushed this I just could not coe up with a proper header

roles/openstackclient/templates/openstack.j2
janhorstmann: I did not have a concrete problem, but thought that, e.g. piping stderr into a file would now result in a lot of non-printable characters in it.
Which issues do you have in mind?

osfrickler: like what happens if someone just redirects stderr, but the command still asks for a password interactively, is this expected to work properly?

osfrickler:
> > Seems like a useful idea, I'm just not sure how to properly test it.
> 
> Maybe I am missing your point, but shouldn't piping the output into something that shows non-printable charcaters do the trick?

I'm not worried about the question whether this change does what it says for interactive use, but that maybe there is some automated use within consumers like testbed that might break. But I think we can just go ahead with this PR and find out.

roles/openstackclient/templates/openstack.j2
janhorstmann: In case of the openstack client it will fall back to a non-interactive mode when no terminal is attached and just print an error that no password ist set. I have not tested any of the other tools though

PR #1802, opened 2025-04-02
Update custom fields in netbox

janhorstmann:
Add:
* out-of-band-address
* out-of-band-port

Remove:
* deployment_enabled
* ironic_enabled
* configuration_template
* device_state
* device_transition
* device_type
* deployment_type
* deployment_state
* introspection_state
* network_interface_name
* ironic_state

Part of https://github.com/osism/issues/issues/1228

### DIFF
--- a/roles/netbox/templates/initializers/custom_fields.yml.j2
+++ b/roles/netbox/templates/initializers/custom_fields.yml.j2
@@ -1,84 +1,4 @@
 ---
-deployment_enabled:
-  type: boolean
-  label: Deployment enabled
-  default: false
-  description: Deployment enabled
-  required: false
-  weight: 0
-  on_objects:
-    - dcim.models.Device
-
-ironic_enabled:
-  type: boolean
-  label: Ironic enabled
-  default: false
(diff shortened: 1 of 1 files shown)


berendt:
@janhorstmann Use oob instead of out-of-band. This way it matches with the prefix labels.

janhorstmann:
> @janhorstmann Use oob instead of out-of-band. This way it matches with the prefix labels.

Done

roles/netbox/templates/initializers/custom_fields.yml.j2
```diff
@@ -109,41 +29,31 @@ power_state:
...
   on_objects:
     - dcim.models.Device
 
-out_of_band_type:
-  type: text
-  label: Out-of-band type
+oob_port:
+  type: integer
+  label: Out-of-band port
   default: ""
```
berendt, line 55: This looks wrong. The default for an integer should not be a string.

berendt (changes requested):

PR #1808, opened 2025-04-15
chrony: use nts by default

berendt:


roles/chrony/defaults/main.yml
```diff
@@ -12,7 +12,7 @@ chrony_servers:
   - 3.de.pool.ntp.org
 
 # NTP server options.
-chrony_server_options: iburst
+chrony_server_options: iburst nts
```
osfrickler, line 15: is this generally supported by most pool servers? or does it fallback to not being used if the server doesn't support it?

berendt: Doesn't look like it. Then we'd better just document it and only add support for custom certificates so that we can use it internally.

jklare:
@berendt @osfrickler is there still work to be done here, or was this moved to be solved in documentation? Was it already documented?

osfrickler:
I'm not aware of any follow-up work having happened. I'm also questioning whether deploying NTS is helpful unless you also deploy a well-designed multi-time-server setup with trusted external time sources, which is a bit off-topic for us. so my suggestion would be to just close this PR

jklare:
I will close this for now then.

PR #1915, opened 2025-06-21
httpd: do not enforce password change on SONiC devices after deployment

berendt:


roles/httpd/templates/post_install.sh.j2
```diff
@@ -4,3 +4,6 @@
 chown -R admin:admin /home/admin/.ssh
 chmod 0700 /home/admin/.ssh
 chmod 0600 /home/admin/.ssh/authorized_keys
+
+# do not enforce a password change after the initial login
+chage -d 0 admin
```
osfrickler, line 9: `-d 99999` seems to work better, moving the next required password change into the year 2243

osfrickler (approved):

PR #1922, opened 2025-06-26
httpd: use /ironic and /sonic

berendt:


roles/httpd/templates/httpd.conf.j2
```diff
@@ -305,6 +300,15 @@ Alias /sonic /usr/local/apache2/htdocs/sonic
 </Directory>
 {% endif %}
 
+{% if httpd_ironic_enable | bool %}
+Alias /ironic /usr/local/apache2/htdocs/ironic
+<Directory "/usr/local/apache2/htdocs/ironic">
```
janhorstmann, line 305: I think we only need/should expose the `httpboot` directory, as is [currently done in the `ironic_http` container](https://opendev.org/openstack/kolla-ansible/src/branch/master/ansible/roles/ironic/templates/ironic-http-httpd.conf.j2#L10)
```suggestion
Alias /ironic /usr/local/apache2/htdocs/ironic/httpboot
<Directory "/usr/local/apache2/htdocs/ironic/httpboot">
```

PR #1999, opened 2025-09-21
manager: add container version verification system

berendt:
- Add Jinja2 template for dynamic version checking script
- Include conditional execution based on service settings
- Verify all containers match expected versions after deployment
- Skip verification when manual start or restart disabled

AI-assisted: Claude Code


berendt:
Recheck

PR #2022, opened 2025-11-07
frr: Add optional BFD support for FRR neighbors

berendt:
Add optional BFD (Bidirectional Forwarding Detection) configuration for BGP neighbors in FRR role.

- Add frr_bfd_profiles variable to defaults/main.yml with example configuration
- Update all FRR templates to support BFD profile configuration
- Add per-neighbor BFD enablement via 'bfd' or 'bfd_profile' attributes
- Support both interface-based and IP-based BGP neighbors
- Maintain backward compatibility - BFD is completely optional

### DIFF
--- a/roles/frr/templates/frr_leaf.conf.j2
+++ b/roles/frr/templates/frr_leaf.conf.j2
@@ -27,10 +27,10 @@ router bgp {{ frr_local_as }}
  neighbor {{ item.interface }} password {{ item.password }}
 {% endif %}
 {# Enable BFD for neighbor if configured #}
-{% if item.bfd is defined and item.bfd %}
- neighbor {{ item.interface }} bfd
-{% elif item.bfd_profile is defined %}
+{% if item.bfd_profile is defined %}
  neighbor {{ item.interface }} bfd profile {{ item.bfd_profile }}
+{% elif item.bfd is defined and item.bfd %}
+ neighbor {{ item.interface }} bfd
 {% endif %}
 {% endfor %}
  !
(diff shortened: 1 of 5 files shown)


jklare:
rebased

jklare:
Hi @fzakfeld , IIRC @berendt was working with you on this, so I requested a review from you. Please let me know if my memory fooled me and I should find somebody else for the review.

fzakfeld:
Hey, we don't use or have implemented BFD but I can have a look later

jklare:
@fzakfeld thanks a lot, much appreciated!

roles/frr/templates/frr_leaf.conf.j2
```diff
@@ -13,6 +26,12 @@ router bgp {{ frr_local_as }}
 {% if "password" in item %}
  neighbor {{ item.interface }} password {{ item.password }}
 {% endif %}
+{# Enable BFD for neighbor if configured #}
+{% if item.bfd is defined and item.bfd %}
+ neighbor {{ item.interface }} bfd
+{% elif item.bfd_profile is defined %}
```
fzakfeld: `bfd_profile` is only used if `bfd` is undefined or falsy. It would be better to check for bfd_profile first, in case a neighbor has both `bfd: true` and a bfd_profile set

fzakfeld (changes requested):

roles/frr/templates/frr_leaf.conf.j2
osfrickler: done

osfrickler (approved):
added requested changes, patch lgtm otherwise

PR #2048, opened 2026-02-24
dnsmasq: use bind-dynamic when no interfaces are configured

berendt:
When _dnsmasq_interfaces is empty, use bind-dynamic instead of bind-interfaces to allow dnsmasq to listen on all interfaces dynamically.


janhorstmann:
I think we ended up not using `bind-dynamic` at all. For the case of no specified `bind-interfaces` I think the default mode of binding the wildcard address seems more appropriate

PR #2076, opened 2026-04-09
docker: set default ulimits for nofile

berendt:


roles/docker/defaults/main.yml
```diff
@@ -157,12 +157,11 @@ docker_redhat_repository: "https://download.docker.com/linux/centos/docker-ce.re
...
-#   nofile:
-#     Name: "nofile"
-#     Soft: 262144
-#     Hard: 1048576
-docker_default_ulimits: {}
+docker_default_ulimits:
+  nofile:
+    Name: "nofile"
+    Soft: 65536
+    Hard: 262144
```
osfrickler: I wonder if we should go with the higher defaults directly matching what was present as comment before? or is there a strong reason to be more conservative?

osfrickler (approved):

PR #2083, opened 2026-05-12
Split service start and active-state wait into separate tasks

berendt:
The generic ansible.builtin.service module does not always populate the status dict with ActiveState, which made result["status"]["ActiveState"] fail with "'dict object' has no attribute 'status'". Switch to ansible.builtin.systemd_service for the start/restart and add a dedicated wait task using systemctl is-active so the until-loop actually observes the activating -> active transition.

AI-assisted: Claude Code

### DIFF
--- a/roles/adminer/handlers/main.yml
+++ b/roles/adminer/handlers/main.yml
@@ -1,10 +1,18 @@
 ---
 - name: Restart adminer service
   become: true
-  ansible.builtin.service:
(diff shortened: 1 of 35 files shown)


jklare:
Looking at the code, it seems we are duplicating the same restarts and checks as tasks and handlers. Is this intended? If so, why is this needed?

PR #2084, opened 2026-05-18
Add su_loadbalancer FRR template

fzakfeld:


osfrickler:
I do not understand the usecase for this template, we would need to add some documentation to this role how the deployment scenarios would look like and what configuration is needed. We should likely do this for the existing templates first, but until then I do not see a particular advantage in adding this here, as it is always possible to use custom templates locally in the config repo that you are deploying from.

fzakfeld:
Hey @osfrickler, so @berendt told me to add my own templates here, similar to the `yrzn` named ones. The current `loadbalancer` template for example lacks the ability to set different source IPs depending on the routes (which is implemented here)

Our general network design for this env is not fixed yet (hence the draft), let's talk about this next week and we might find a generic template approach that is not locked to our usecase.

berendt:
@fzakfeld Can you use scaleup instead of su? For me it's fine to an environment specific templates as long as we have a code owner for the file. We should document this so that everybody understands that yrzn and scaleup are environment specific configurations that should not be used on other environments.

fzakfeld:
@berendt yes, that would work. But let's first talk about our design next week, perhaps we can agree on some generic config overall

PR #2090, opened 2026-06-01
molecule: publish per-container logs on failure

ideaship:
## What

A Zuul `post-run` playbook (`playbooks/collect-container-logs.yml`), wired
into `abstract-ansible-collection-services-molecule`, that publishes
per-container docker diagnostics on every molecule run into the build's log
directory (per node):

- `_overview.txt` — `docker ps --all` (name / state / status)
- `<container>.txt` — `docker logs` (stdout+stderr)
- `<container>.inspect.json` — `docker inspect .State` (health / restart / exit)

Best-effort (`failed_when: false`); covers every docker-based role in the
scenario.

## Why

The molecule jobs published nothing about the running containers, so a
container that never became healthy could only be diagnosed by a local
reproduction. This is the gap that left the chronic
`molecule-netbox-worker-unhealthy` flake (~70% failure rate) unexplained
for months. With this instrumentation, the netbox flake was root-caused
from a single CI failure — the fix is in #2091.

A `post-run` is required rather than a `block`/`rescue` in the converge
playbook: the molecule health gates run as Ansible *handlers*, whose
failures happen at play-end (`flush_handlers`) outside any block/rescue, so
a rescue never fires. The post phase runs after molecule aborts, while the
containers are still up (`destroy` is off by default).

## Safety

Molecule runs use dummy credentials, and `docker logs` captures container
stdout/stderr (not the environment), so no real secrets are published.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -80,6 +80,7 @@
     name: abstract-ansible-collection-services-molecule
     parent: ansible-molecule
     abstract: true
+    post-run: playbooks/collect-container-logs.yml
     nodeset:
       nodes:
         - name: debian-bookworm
--- a/playbooks/collect-container-logs.yml
+++ b/playbooks/collect-container-logs.yml
@@ -1,0 +1,72 @@
+---
+# Zuul post-run for the molecule jobs: publish per-container docker logs so a
+# container that never becomes healthy (e.g. the chronically flaky netbox
+# rqworker) can be diagnosed from the build logs instead of requiring a local
+# reproduction.
+#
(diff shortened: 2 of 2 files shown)


ideaship:
recheck

ideaship:
recheck

ideaship:
recheck

ideaship:
recheck

ideaship:
recheck

ideaship:
recheck

osfrickler:
recheck

PR #2097, opened 2026-06-16
smartd: make /etc/smartd.conf configurable via variables

### LINKED ISSUE #2082: smartd: smartd role is not configurable for /etc/smartd.conf
## Description

The current `smartd` role is quite static and does not provide a way to manage or customize the actual `smartd` configuration file, especially `/etc/smartd.conf`.

At the moment, the role installs `smartmontools`, manages `/etc/default/smartmontools`, and starts/enables the service. However, there does not seem to be a role variable or template mechanism to modify the devices monitored by `smartd`.

This becomes a problem on systems where the default `DEVICESCAN` configuration does not work.

## Issue

I am running the role on a server with an older `HPE Smart Array Gen8 Controllers` RAID controller.

On this system, `smartd` is not able to detect any disks through the default `DEVICESCAN` configuration.

The service fails with:

```text
May 11 12:39:06 controller-01 smartd[70641]: smartd 7.4 2023-08-01 r5530 [x86_64-linux-6.8.0-111-generic] (local build)
May 11 12:39:06 controller-01 smartd[70641]: Opened configuration file /etc/smartd.conf
May 11 12:39:06 controller-01 smartd[70641]: Drive: DEVICESCAN, implied '-a' Directive on line 18 of file /etc/smartd.conf
May 11 12:39:06 controller-01 smartd[70641]: Configuration file /etc/smartd.conf was parsed, found DEVICESCAN, scanning devices
May 11 12:39:06 controller-01 smartd[70641]: In the system's table of devices NO devices found to scan
May 11 12:39:06 controller-01 smartd[70641]: Unable to monitor any SMART enabled devices. Exiting.
May 11 12:39:06 controller-01 smartd[70641]: smartd is exiting (exit status 17)
May 11 12:39:06 controller-01 systemd[1]: smartmontools.service: Main process exited, code=exited, status=17/n/a
```

The Ansible task then fails when trying to start the service:

```text
TASK [osism.services.smartd : Manage smartd service] ***************************
fatal: [ost-controller-17]: FAILED! => {"changed": false, "msg": "Unable to start service smartd: Job for smartmontools.service failed because the control process exited with error code.\nSee \"systemctl status smartmontools.service\" and \"journalctl -xeu smartmontools.service\" for details.\n"}
```

## Manual Workaround

During debugging, I found that the disks behind the RAID controller can be queried with:

```bash
sudo smartctl -a /dev/sda -d cciss,0
```

As a workaround, I manually edited `/etc/smartd.conf` and replaced the default `DEVICESCAN` configuration:

```text
#DEVICESCAN -d removable -n standby -m root -M exec /usr/share/smartmontools/smartd-runner
/dev/sda -d cciss,0 -a -n standby -m root -M exec /usr/share/smartmontools/smartd-runner
```

After this manual change, `smartd` can monitor the disk correctly.

## Requested Improvement

It would be very helpful if the `smartd` role allowed managing `/etc/smartd.conf` through Ansible variables.

For example, the role could support something like:

```yaml
smartd_devices:
  - device: /dev/sda
    type: cciss,0
    options: "-a -n standby -m root -M exec /usr/share/smartmontools/smartd-runner"
```

It would also be useful to have a way to disable or replace the default `DEVICESCAN` line.

berendt:
Closes #2082

## Problem

The `smartd` role installs `smartmontools`, manages
`/etc/default/smartmontools`, and starts the service, but never manages
`/etc/smartd.conf`. As a result the package-provided `DEVICESCAN` line
cannot be replaced when it finds no devices — e.g. behind an HPE Smart
Array Gen8 RAID controller, where the daemon exits with status 17
(`Unable to monitor any SMART enabled devices`) and the
`Manage smartd service` task fails.

## Change set (in commit order)

1. **`smartd: make /etc/smartd.conf configurable via variables`**
   - `roles/smartd/templates/smartd.conf.j2` (new) — renders one line
     per device as `<device> -d <type> <options>` (type/options
     optional), or a `DEVICESCAN` line when no devices are configured.
   - `roles/smartd/tasks/main.yml` — new `Copy smartd configuration
     file` task (`ansible.builtin.template` → `/etc/smartd.conf`,
     `0644 root:root`, `notify: Restart smartd service`), inserted
     between the existing `/etc/default/smartmontools` copy and
     `Manage smartd service`. Gated by
     `when: smartd_configure | bool or smartd_devices | length > 0`.
   - `roles/smartd/defaults/main.yml` — documents the new variables:
     - `smartd_configure` (default `false`) — gate; also auto-enables
       when `smartd_devices` is set.
     - `smartd_devices` (default `[]`) — list of `{device, type?,
       options?}`; setting any device replaces the default
       `DEVICESCAN` line.
     - `smartd_devicescan` (default `true`) — emit `DEVICESCAN` when no
       devices are configured.
     - `smartd_devicescan_options` — directives appended to
       `DEVICESCAN` (default matches the Debian/Ubuntu package default).

2. **`smartd: cover /etc/smartd.conf rendering in molecule test`**
   - `molecule/delegated/vars/smartd.yml` — sets `smartd_devices` to the
     issue's example so converge exercises the new path.
   - `molecule/delegated/tests/smartd.py` — `test_smartd_conf` asserts
     `/etc/smartd.conf` is created `root:root 0644`, renders the device
     line verbatim, and contains no `DEVICESCAN` line.

## Why opt-in by default

Management defaults to **off** so existing Debian/Ubuntu/EL9 hosts and
any hand-edited `/etc/smartd.conf` (including the manual workaround from
the issue) are left untouched until a user opts in — either by setting
`smartd_devices` or `smartd_configure: true`. This also avoids a
potential EL9 regression: `smartd_devicescan_options` defaults to the
Debian path `/usr/share/smartmontools/smartd-runner`, which differs on
RedHat-family systems; keeping management opt-in means the default has no
effect there.

## Verification

Rendered the template via a throwaway localhost play for all branches
(devices set, DEVICESCAN on, DEVICESCAN off, device without type/options)
— the devices case reproduces the issue's workaround line exactly with a
single trailing newline and no `DEVICESCAN`. Local checks run green:
`yamllint roles/smartd`, `ansible-lint roles/smartd` (profile
`production`, 0 findings), `flake8` + `black --check` on the test.

Note: the `delegated` molecule scenario is `managed: false` (remote
host), so the full converge/verify runs in the Zuul
`ansible-collection-services-molecule-smartd` job. The
`smartd_configure: true` + empty-`smartd_devices` DEVICESCAN path is
covered by template review only — the scenario can converge a single
variable set.

### DIFF
--- a/molecule/delegated/tests/smartd.py
+++ b/molecule/delegated/tests/smartd.py
@@ -44,6 +44,17 @@ def test_smartd_conf(host):
     assert "DEVICESCAN" not in config_file.content_string
 
 
+def test_smartd_conf_devicescan(host):
+    config_file = host.file("/etc/smartd.conf.devicescan")
+    assert config_file.exists
+    assert config_file.is_file
+    assert (
+        "DEVICESCAN -d removable -n standby -m root "
+        "-M exec /usr/share/smartmontools/smartd-runner" in config_file.content_string
+    )
+    assert "/dev/sda" not in config_file.content_string
+
+
 """
 The ConditionVirtualization is a systemd condition that checks whether the system is
 running in a virtualized environment. It can be used as part of the systemd service
--- a/molecule/delegated/tests/smartd.py
+++ b/molecule/delegated/tests/smartd.py
@@ -30,6 +30,20 @@ def test_smartmontools_config(host):
     )
 
 
+def test_smartd_conf(host):
+    config_file = host.file("/etc/smartd.conf")
+    assert config_file.exists
+    assert config_file.is_file
+    assert config_file.user == "root"
+    assert config_file.group == "root"
+    assert config_file.mode == 0o644
+    assert (
+        "/dev/sda -d cciss,0 -a -n standby -m root "
+        "-M exec /usr/share/smartmontools/smartd-runner" in config_file.content_string
+    )
+    assert "DEVICESCAN" not in config_file.content_string
+
+
 """
 The ConditionVirtualization is a systemd condition that checks whether the system is
 running in a virtualized environment. It can be used as part of the systemd service
--- a/roles/smartd/defaults/main.yml
+++ b/roles/smartd/defaults/main.yml
@@ -4,3 +4,29 @@
 
 smartd_package_name: smartmontools
 smartd_service_name: smartd
+
+##########################
+# configuration
+
+# Manage /etc/smartd.conf. When false the file is left untouched, which
+# preserves the package default and any manual edits. Management is also
+# enabled automatically when smartd_devices is not empty.
+smartd_configure: false
+
+# Devices to monitor. Each entry renders one line in /etc/smartd.conf as
+# "<device> -d <type> <options>". The type and options keys are optional.
+# Setting any device replaces the default DEVICESCAN line.
+# Example:
+#   smartd_devices:
+#     - device: /dev/sda
+#       type: cciss,0
+#       options: "-a -n standby -m root -M exec /usr/share/smartmontools/smartd-runner"
+smartd_devices: []
+
+# Emit a DEVICESCAN line when no smartd_devices are configured. Set to false
+# to disable the default device scan. Ignored when smartd_devices is set.
+smartd_devicescan: true
+
+# Directives appended to the DEVICESCAN line. The default matches the
+# Debian/Ubuntu smartmontools package default.
+smartd_devicescan_options: "-d removable -n standby -m root -M exec /usr/share/smartmontools/smartd-runner"
--- a/roles/smartd/defaults/main.yml
+++ b/roles/smartd/defaults/main.yml
@@ -23,8 +23,11 @@ smartd_configure: false
 #       options: "-a -n standby -m root -M exec /usr/share/smartmontools/smartd-runner"
 smartd_devices: []
 
-# Emit a DEVICESCAN line when no smartd_devices are configured. Set to false
-# to disable the default device scan. Ignored when smartd_devices is set.
+# Emit a DEVICESCAN line when no smartd_devices are configured. Only takes
+# effect when /etc/smartd.conf is managed (smartd_configure: true); with the
+# default smartd_configure: false the file is left untouched and the package's
+# own DEVICESCAN line stays active regardless of this setting. Set to false to
+# suppress the DEVICESCAN line. Ignored when smartd_devices is set.
 smartd_devicescan: true
 
 # Directives appended to the DEVICESCAN line. The default matches the
--- a/roles/smartd/defaults/main.yml
+++ b/roles/smartd/defaults/main.yml
@@ -20,7 +20,7 @@ smartd_configure: false
 #   smartd_devices:
 #     - device: /dev/sda
 #       type: cciss,0
-#       options: "-a -n standby -m root -M exec /usr/share/smartmontools/smartd-runner"
+#       options: "-a -n standby -m root"
 smartd_devices: []
 
 # Emit a DEVICESCAN line when no smartd_devices are configured. Only takes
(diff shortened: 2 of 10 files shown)


roles/smartd/defaults/main.yml
```diff
@@ -4,3 +4,29 @@
...
+#       options: "-a -n standby -m root -M exec /usr/share/smartmontools/smartd-runner"
+smartd_devices: []
+
+# Emit a DEVICESCAN line when no smartd_devices are configured. Set to false
+# to disable the default device scan. Ignored when smartd_devices is set.
+smartd_devicescan: true
+
+# Directives appended to the DEVICESCAN line. The default matches the
+# Debian/Ubuntu smartmontools package default.
+smartd_devicescan_options: "-d removable -n standby -m root -M exec /usr/share/smartmontools/smartd-runner"
```
ideaship: `meta/main.yml` advertises EL/RHEL support, but this default points `-M exec` at `/usr/share/smartmontools/smartd-runner` — a path that exists only on Debian/Ubuntu, not on RHEL. An EL operator who opts in (`smartd_configure: true` with empty `smartd_devices` is a supported mode) gets a managed `/etc/smartd.conf` referencing a helper that isn't installed, so device-failure alerts silently never fire. This widens an existing drift away from the role's advertised RHEL support: it already writes the Debian-only `/etc/default/smartmontools` unconditionally (`tasks/main.yml:14-22`).

roles/smartd/defaults/main.yml
```diff
@@ -4,3 +4,29 @@
...
+# "<device> -d <type> <options>". The type and options keys are optional.
+# Setting any device replaces the default DEVICESCAN line.
+# Example:
+#   smartd_devices:
+#     - device: /dev/sda
+#       type: cciss,0
+#       options: "-a -n standby -m root -M exec /usr/share/smartmontools/smartd-runner"
+smartd_devices: []
+
+# Emit a DEVICESCAN line when no smartd_devices are configured. Set to false
```
ideaship: `smartd_configure` defaults to `false`, so in the default configuration this comment is incorrect: setting only `smartd_devicescan: false` does nothing. The render task (`tasks/main.yml:32`) is gated on `smartd_configure | bool or smartd_devices | length > 0`, so with config management off (the default) the file is never written and the package's own `DEVICESCAN` stays active.

roles/smartd/templates/smartd.conf.j2
```diff
@@ -0,0 +1,7 @@
+# {{ ansible_managed }}
+{% for device in smartd_devices %}
+{{ device.device }}{{ (' -d ' + device.type) if device.type is defined else '' }}{{ (' ' + device.options) if device.options is defined else '' }}
```
ideaship: Use truthiness, not just `is defined`, so an empty string doesn't emit a dangling flag: `... if device.type is defined and device.type else ''` (same for `device.options`). As written, `type: ""` renders `-d ` with no argument and smartd will reject the file.

molecule/delegated/tests/smartd.py
```diff
@@ -30,6 +30,20 @@ def test_smartmontools_config(host):
     )
 
 
+def test_smartd_conf(host):
```
ideaship, line 33: `test_smartd_conf` only covers the explicit-devices branch. The DEVICESCAN-default path (`smartd_configure: true`, empty `smartd_devices`) is the scenario the commit message motivates (replacing the package DEVICESCAN behind a RAID controller) and is currently untested. Consider a follow-up that exercises the empty-devices render and asserts the rendered `DEVICESCAN` line.

ideaship (changes requested):
Consider adding `argument_specs.yml` now that the role takes a structured `smartd_devices` list. Requiring `device` and typing `type`/`options` as strings would turn the current cryptic Jinja `'dict object' has no attribute 'device'` into a clear validation error, and is the practical substitute for a config-validate hook.

ideaship (approved):

PR #2112, opened 2026-06-17
Add optional prometheus-frr-exporter service to the FRR role

### LINKED ISSUE #2111: Add optional prometheus-frr-exporter service to the FRR role
**Category**: feature | **Scope**: Medium

## Description

The FRR role gains an optional prometheus-frr-exporter service that exposes metrics from the local FRR instance. The exporter is off by default and deployed only when explicitly enabled. Operators can set the listen address and listen port; the default binds to localhost, following how other roles expose local-only endpoints.

The scope is the exporter endpoint itself, nothing more. Wiring the metrics into an existing Prometheus or monitoring stack is out of scope. The exporter should run on Ubuntu and Debian, where the package (`prometheus-frr-exporter` on Ubuntu) is expected to be the same. EL9 support is welcome but optional — if no suitable package exists there, skipping EL9 is acceptable.

## Motivation

Without a metrics endpoint there is no way to monitor or alert on FRR state such as BGP and OSPF session health. Operators who run FRR through this role have no built-in visibility into the routing daemon, so a dropped session or a flapping peer stays silent until something downstream breaks. The exporter gives them a metrics source to scrape with the Prometheus stack OSISM already runs, turning routing failures into observable, alertable signals. Keeping it off by default and bound to localhost means existing deployments are unaffected unless an operator opts in.

---

_Drafted by [planwerk-review](https://github.com/planwerk/planwerk-review) with Claude_

berendt:
Closes #2111

## Summary

Adds an optional `prometheus-frr-exporter` service to the `frr` role.
The exporter exposes metrics from the local FRR instance (BGP/OSPF/BFD
session health and more). It is **off by default** and deployed only on
Debian-family hosts when an operator explicitly enables it. The listen
endpoint binds to `127.0.0.1:9342` by default, following how other
roles expose local-only endpoints (e.g. `scaphandre_host`/`_port`).

Scope is the exporter endpoint itself — wiring the metrics into a
Prometheus/monitoring stack is out of scope, per the issue.

## Change set (in commit order)

**1. `Add optional prometheus-frr-exporter service to frr role`**
- `roles/frr/defaults/main.yml` — new variables: `frr_exporter_enable`
  (`false`), `frr_exporter_package_name`, `frr_exporter_service_name`,
  `frr_exporter_host` (`127.0.0.1`), `frr_exporter_port` (`9342`).
- `roles/frr/tasks/exporter.yml` — installs the package (`apt`), renders
  `/etc/default/prometheus-frr-exporter`, and manages the service.
  Modeled on the in-repo `smartd` role.
- `roles/frr/templates/prometheus-frr-exporter.j2` — renders
  `ARGS="--web.listen-address={{ frr_exporter_host }}:{{ frr_exporter_port }}"`,
  the `$ARGS` variable the packaged systemd unit sources.
- `roles/frr/tasks/main.yml` — includes `exporter.yml` as the final
  task, gated on `frr_exporter_enable | bool` and
  `ansible_os_family == "Debian"`.
- `roles/frr/handlers/main.yml` — new `Restart prometheus-frr-exporter
  service` handler.

**2. `Add molecule coverage for prometheus-frr-exporter`**
- `molecule/delegated/vars/frr.yml` — enables the exporter in the frr
  scenario so CI (debian-bookworm, ubuntu-noble) exercises it.
- `molecule/delegated/tests/frr.py` — four testinfra checks (package
  installed, defaults file rendered with the expected listen-address,
  service running + enabled, endpoint listening), each skipping when
  `frr_exporter_enable` is falsey — mirroring the existing daemon-gated
  tests.

## Platform support

- **Ubuntu / Debian**: the `prometheus-frr-exporter` package is present
  in Debian bookworm (`1.1.4-2`) and Ubuntu noble `universe` (`1.2.0`).
  The packaged systemd unit runs as `frr:frrvty`, so it already has the
  FRR socket access needed to populate metrics — no extra wiring.
- **EL9**: intentionally skipped (no upstream package). Enabling
  `frr_exporter_enable` on EL9 is a deliberate, documented no-op via the
  `ansible_os_family == "Debian"` guard, matching the issue's "skipping
  EL9 is acceptable."

## Design notes

- The new restart handler is intentionally **not** gated on
  `frr_allow_service_restart`. That toggle exists to avoid bouncing the
  routing daemon and dropping BGP/OSPF sessions; restarting the metrics
  exporter has no routing impact, and the config change must take
  effect. If maintainers prefer it to honor the toggle, adding
  `when: frr_allow_service_restart | bool` is a one-line change.

## Verification

- `yamllint` — pass (role + molecule vars)
- `ansible-lint roles/frr` — pass (0 failures, production profile)
- `flake8 molecule/delegated/tests/frr.py` — pass
- Full molecule integration (`delegated` scenario) runs in CI via
  `ansible-collection-services-molecule-frr`; it requires a
  Debian/Ubuntu host and cannot run on the macOS dev box.

---

_Drafted by [planwerk-review](https://github.com/planwerk/planwerk-review) with Claude:claude-opus-4-8_

### DIFF
--- a/molecule/delegated/tests/frr.py
+++ b/molecule/delegated/tests/frr.py
@@ -126,4 +126,23 @@ def test_exporter_listening(host):
 
     host_addr = get_variable(host, "frr_exporter_host")
     port = get_variable(host, "frr_exporter_port")
+
+    # The exporter must be reachable on its configured address ...
     assert host.socket(f"tcp://{host_addr}:{port}").is_listening
+
+    # ... and bound *only* there. is_listening against a wildcard bind
+    # (0.0.0.0/::) also succeeds for loopback connections, so it would pass even
+    # if the exporter ignored /etc/default/... and fell back to 0.0.0.0:9342.
+    # Assert the address actually in effect matches frr_exporter_host exactly,
+    # which both rejects a wildcard exposure and proves the rendered
+    # --web.listen-address argument took effect. ss reports IPv6 literals
+    # bracketed, mirroring the template.
+    expected = f"[{host_addr}]:{port}" if ":" in host_addr else f"{host_addr}:{port}"
+    local_addrs = [
+        line.split()[3]
+        for line in host.check_output(f"ss -ltnH 'sport = :{port}'").splitlines()
+        if line.strip()
+    ]
+    assert local_addrs, f"nothing is listening on port {port}"
+    for addr in local_addrs:
+        assert addr == expected, f"exporter bound to {addr}, expected {expected}"
--- a/molecule/delegated/tests/frr.py
+++ b/molecule/delegated/tests/frr.py
@@ -84,3 +84,46 @@ def test_function_bfdd(host):
         result = host.run("vtysh -c 'show bfd peers'")
         assert result.rc == 0
         assert "BFD Peers" in result.stdout
+
+
+def test_exporter_pkg(host):
+    if not get_variable(host, "frr_exporter_enable"):
+        pytest.skip("frr_exporter_enable disabled")
+
+    package = host.package(get_variable(host, "frr_exporter_package_name"))
+    assert package.is_installed
+
+
+def test_exporter_defaultsfile(host):
+    if not get_variable(host, "frr_exporter_enable"):
+        pytest.skip("frr_exporter_enable disabled")
+
+    service_name = get_variable(host, "frr_exporter_service_name")
+    host_addr = get_variable(host, "frr_exporter_host")
+    port = get_variable(host, "frr_exporter_port")
+
+    f = host.file(f"/etc/default/{service_name}")
+    assert f.exists
+    assert not f.is_directory
+    assert f.mode == 0o644
+    assert f.user == "root"
+    assert f.group == "root"
+    assert f"--web.listen-address={host_addr}:{port}" in f.content_string
+
+
+def test_exporter_srv(host):
+    if not get_variable(host, "frr_exporter_enable"):
+        pytest.skip("frr_exporter_enable disabled")
+
+    service = host.service(get_variable(host, "frr_exporter_service_name"))
+    assert service.is_running
+    assert service.is_enabled
+
+
+def test_exporter_listening(host):
+    if not get_variable(host, "frr_exporter_enable"):
+        pytest.skip("frr_exporter_enable disabled")
+
+    host_addr = get_variable(host, "frr_exporter_host")
+    port = get_variable(host, "frr_exporter_port")
+    assert host.socket(f"tcp://{host_addr}:{port}").is_listening
--- a/roles/frr/handlers/main.yml
+++ b/roles/frr/handlers/main.yml
@@ -6,6 +6,10 @@
     state: restarted
   when: frr_allow_service_restart | bool
 
+# Unlike "Restart frr service", this restart is intentionally not gated on
+# frr_allow_service_restart: bouncing the exporter is data-plane-safe (it does
+# not touch FRR routing) and we want bind-address changes (e.g. tightening
+# frr_exporter_host from a wildcard back to localhost) to take effect at once.
 - name: Restart prometheus-frr-exporter service
   become: true
   ansible.builtin.service:
--- a/roles/frr/handlers/main.yml
+++ b/roles/frr/handlers/main.yml
@@ -5,3 +5,9 @@
     name: "{{ frr_service_name }}"
     state: restarted
   when: frr_allow_service_restart | bool
+
+- name: Restart prometheus-frr-exporter service
+  become: true
+  ansible.builtin.service:
+    name: "{{ frr_exporter_service_name }}"
+    state: restarted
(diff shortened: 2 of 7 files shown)


roles/frr/tasks/main.yml
```diff
@@ -110,3 +110,9 @@
     name: "{{ frr_service_name }}"
     state: started
     enabled: true
+
+- name: Include prometheus-frr-exporter tasks
```
ideaship, line 114: There's currently no way to turn this exporter off: `frr_exporter_enable: false` doesn't disable it, it just makes the role stop managing it. This include is the only thing that touches the exporter, and it's gated on `frr_exporter_enable | bool` — so once an operator enables it (package installed, service started/enabled, defaults file written), flipping the flag back to `false` does *nothing*. The package stays installed and the service keeps running and listening. That contradicts the `defaults/main.yml` comment ("only deployed ... when explicitly enabled"), and it's more pointed if someone set `frr_exporter_host: 0.0.0.0`, since the endpoint stays exposed after they think they turned it off.

Please either **document** that the flag is install-only and does not stop an already-deployed exporter, or **change it** to actually tear down.

roles/frr/handlers/main.yml
```diff
@@ -5,3 +5,9 @@
     name: "{{ frr_service_name }}"
     state: restarted
   when: frr_allow_service_restart | bool
+
+- name: Restart prometheus-frr-exporter service
```
ideaship, line 13: The existing `Restart frr service` handler above respects `when: frr_allow_service_restart | bool`, but this new handler restarts unconditionally, so an operator who set `frr_allow_service_restart: false` to suppress mid-run restarts will still get the exporter bounced. An exporter restart is data-plane-safe, so exempting it is reasonable — this is just a request to make it deliberate: either add `when: frr_allow_service_restart | bool` for consistency, or leave it unconditional with a one-line comment noting the exemption is intentional.

molecule/delegated/tests/frr.py
```diff
@@ -84,3 +84,46 @@ def test_function_bfdd(host):
...
+    assert service.is_enabled
+
+
+def test_exporter_listening(host):
+    if not get_variable(host, "frr_exporter_enable"):
+        pytest.skip("frr_exporter_enable disabled")
+
+    host_addr = get_variable(host, "frr_exporter_host")
+    port = get_variable(host, "frr_exporter_port")
+    assert host.socket(f"tcp://{host_addr}:{port}").is_listening
```
ideaship, line 131: `host.socket("tcp://127.0.0.1:9342").is_listening` is satisfied by a wildcard `0.0.0.0:9342` bind too, since loopback connections succeed against a wildcard listener. So this asserts "reachable on localhost," not "bound only to localhost" — which is the property the defaults comment actually promises. If the package ever ignored `/etc/default/...` and fell back to its built-in `0.0.0.0:9342` default, this whole suite would still pass while the host was exposed. Please add a negative assertion — `assert not host.socket("tcp://0.0.0.0:9342").is_listening` — or parse `ss -ltnH 'sport = :9342'` and assert the bound address is exactly `127.0.0.1`.

roles/frr/tasks/exporter.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Install prometheus-frr-exporter package
+  become: true
+  ansible.builtin.apt:
+    name: "{{ frr_exporter_package_name }}"
+    state: present
+    lock_timeout: "{{ apt_lock_timeout | default(300) }}"
+
+- name: "Copy file: /etc/default/{{ frr_exporter_service_name }}"
```
ideaship: This file only controls the exporter's bind address because the Debian package's systemd unit reads it via `EnvironmentFile=/etc/default/prometheus-frr-exporter` and passes `$ARGS` to `ExecStart`. That's the standard Debian exporter convention and correct today, but it's an invisible contract: if a package revision renames the file or drops `$ARGS`, this template is silently ignored and the exporter falls back to `0.0.0.0:9342` while the role still reports success. Please add a one-line comment here recording the `EnvironmentFile`/`$ARGS` dependency — and pairing this with the strengthened bind-address test above would actually prove the rendered argument took effect.

roles/frr/templates/prometheus-frr-exporter.j2
```diff
@@ -0,0 +1,3 @@
+# Command-line arguments for the prometheus-frr-exporter service.
+# Managed by Ansible (osism.services.frr role).
+ARGS="--web.listen-address={{ frr_exporter_host }}:{{ frr_exporter_port }}"
```
ideaship: `{{ frr_exporter_host }}:{{ frr_exporter_port }}` isn't safe for IPv6 literals. `frr_exporter_host: ::1` renders `--web.listen-address=::1:9342`, which Go's `net.SplitHostPort` rejects — IPv6 needs `[::1]:9342`, so the exporter won't start. Either bracket conditionally — `{{ ('[' ~ frr_exporter_host ~ ']') if ':' in frr_exporter_host else frr_exporter_host }}` — or document in `defaults/main.yml` that `frr_exporter_host` is IPv4/hostname-only.

ideaship (changes requested):

ideaship (approved):

PR #2114, opened 2026-06-17
squid: add support for network host mode

### LINKED ISSUE #2113: Run squid with host networking
As of now, Docker container networking has problems with routed networking setups, where a manual outbound NAT rule is needed, e.g.

```
iptables -t nat -I POSTROUTING -s 172.31.100.0/23 -d 10.141.167.0/24 -j SNAT --to-source 10.141.167.10
```

For squid, this would mean creating such a rule for 0.0.0.0/0. With complex routing tables, we have to mimic the source addresses of the routing table. It would be much easier to run squid with host networking.

I will add this as an option, the issue is just to track it

berendt:
## Summary

Docker bridge networking requires manual outbound SNAT rules in routed
setups. For squid this means maintaining a SNAT rule for `0.0.0.0/0`
that has to mirror the source addresses of (potentially complex)
routing tables. Running squid with Docker **host networking** sidesteps
the bridge entirely and removes that burden.

This PR adds an optional `squid_network_mode` variable (default
`bridge`). Setting it to `host` runs the squid container with
`network_mode: host`. The change is purely additive Jinja conditionals
in one template plus one new default and mirrors the existing
`dnsdist_network_mode` (#2074) and `httpd_network_mode` options already
in this repo.

Closes #2113

## Changes (in commit order)

**`squid: add support for network host mode`**

- `roles/squid/defaults/main.yml` — add `squid_network_mode: bridge`
  (placed next to `squid_network`, matching the dnsdist role layout).
- `roles/squid/templates/docker-compose.yml.j2` — two surgical
  conditionals:
  - In `host` mode emit `network_mode: host`; otherwise keep the
    existing `ports:` mapping (`{{ squid_host }}:{{ squid_port }}:3128`).
  - Emit the `networks:` bridge block only when not in `host` mode.
  - The `volumes:\n  squid:` block is retained in **both** modes.

## Verification

- Rendered the template with Ansible's Jinja2 (`trim_blocks=True`):
  - `squid_network_mode: bridge` → **byte-identical** to the current
    template output, so existing deployments are unaffected.
  - `squid_network_mode: host` → emits `network_mode: host`, omits
    `ports:` and the `networks:` block, and keeps `volumes:\n  squid:`.
- `yamllint -c .yamllint.yml roles/squid/defaults/main.yml` — passes.
- The existing `molecule/delegated` scenario
  (`molecule/delegated/tests/squid.py`) continues to exercise the
  default `bridge` path unchanged; it is the regression guard for
  "default behavior is untouched". It runs in Zuul CI against a
  delegated host.

## Security note

In `bridge` mode squid is reachable only at `squid_host:squid_port`
(default `127.0.0.1:3128`). In `host` mode the Docker port mapping no
longer applies and squid listens on the base image default
`http_port 3128` on all host interfaces, with `http_access allow all`
from the role's `conf.d` drop-in. Host mode is therefore an explicit
operator opt-in (the default stays `bridge` + `127.0.0.1`, so
secure-by-default is preserved) and operators must firewall the proxy
at the host. The role cannot restrict the bind from a `conf.d` drop-in
without templating the base `squid.conf`, which is out of scope for
this issue.

## Notes

- No documentation changes: `roles/squid/README.md` is empty and
  `CHANGELOG.md` is curated at release time (the dnsdist #2074 entry is
  likewise absent), matching the precedent set by #2074.
- No new automated test, matching the dnsdist precedent (`7cafb656`).

---

_Drafted by [planwerk-review](https://github.com/planwerk/planwerk-review) e69a237 with Claude:claude-opus-4-8_

### DIFF
--- a/roles/squid/templates/docker-compose.yml.j2
+++ b/roles/squid/templates/docker-compose.yml.j2
@@ -11,6 +11,7 @@ services:
       - "{{ squid_host | ansible.utils.ipwrap }}:{{ squid_port }}:3128"
 {% endif %}
     volumes:
+      - "{{ squid_configuration_directory }}/squid.conf:/etc/squid/squid.conf:ro"
       - "{{ squid_configuration_directory }}/conf.d:/etc/squid/conf.d:ro"
       - squid:/var/spool/squid
     healthcheck:
--- a/roles/squid/templates/docker-compose.yml.j2
+++ b/roles/squid/templates/docker-compose.yml.j2
@@ -4,13 +4,18 @@ services:
     container_name: "{{ squid_container_name }}"
     restart: unless-stopped
     image: "{{ squid_image }}"
+{% if squid_network_mode == "host" %}
+    network_mode: host
+{% else %}
     ports:
       - "{{ squid_host | ansible.utils.ipwrap }}:{{ squid_port }}:3128"
+{% endif %}
     volumes:
       - "{{ squid_configuration_directory }}/conf.d:/etc/squid/conf.d:ro"
       - squid:/var/spool/squid
     healthcheck:
       test: pgrep -f '/usr/sbin/squid -f /etc/squid/squid.conf -NYC'
+{% if squid_network_mode != "host" %}
 
 networks:
   default:
@@ -21,6 +26,7 @@ networks:
       driver: default
       config:
         - subnet: {{ squid_network }}
+{% endif %}
 
 volumes:
   squid:
--- a/roles/squid/templates/osism.conf.j2
+++ b/roles/squid/templates/osism.conf.j2
@@ -7,3 +7,7 @@ max_filedescriptors 1048576
 cache_dir ufs /var/spool/squid 100 16 256
 
 http_access allow all
+{% if squid_network_mode == "host" %}
+
+http_port {{ squid_host | ansible.utils.ipwrap }}:{{ squid_port }}
+{% endif %}
--- a/roles/squid/templates/osism.conf.j2
+++ b/roles/squid/templates/osism.conf.j2
@@ -7,7 +7,3 @@ max_filedescriptors 1048576
 cache_dir ufs /var/spool/squid 100 16 256
 
 http_access allow all
-{% if squid_network_mode == "host" %}
-
-http_port {{ squid_host | ansible.utils.ipwrap }}:{{ squid_port }}
-{% endif %}
--- a/molecule/delegated/tests/squid.py
+++ b/molecule/delegated/tests/squid.py
@@ -1,6 +1,37 @@
+import os
+import re
+
 from .util.util import get_ansible, get_variable
 
 testinfra_runner, testinfra_hosts = get_ansible()
+
+ROLE_DIR = os.path.abspath(
+    os.path.join(os.path.dirname(__file__), "..", "..", "..", "roles", "squid")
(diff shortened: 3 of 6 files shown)


roles/squid/templates/osism.conf.j2
```diff
@@ -7,3 +7,7 @@ max_filedescriptors 1048576
 cache_dir ufs /var/spool/squid 100 16 256
 
 http_access allow all
+{% if squid_network_mode == "host" %}
+
+http_port {{ squid_host | ansible.utils.ipwrap }}:{{ squid_port }}
```
ideaship: **Blocking:** this adds a *second* `http_port` rather than replacing the image's default. The role only mounts `conf.d/` (`docker-compose.yml.j2:14`) and squid still loads its stock `/etc/squid/squid.conf`, which defines `http_port 3128` on `0.0.0.0`. Confirmed directly against the image:

```
$ docker run --rm --entrypoint cat ubuntu/squid:6.1-23.10_beta /etc/squid/squid.conf | grep -n '^http_port'
2155:http_port 3128
```

Result in host mode:

- With the default `squid_port: 3128`, squid gets `0.0.0.0:3128` **and** `127.0.0.1:3128` → second `bind()` returns `EADDRINUSE`, which squid treats as FATAL. Host mode won't start out of the box.
- With a non-default port, the stock `0.0.0.0:3128` listener stays up — combined with `http_access allow all` (line 9) that's an open proxy on every host interface, the opposite of the intended confinement.

`conf.d` can't remove a cumulative `http_port`, so adding one here isn't enough. The role needs to manage the base `squid.conf` (mount a templated file that sets `http_port` conditionally and `include`s `conf.d`). The cited precedents (`dnsdist`, `httpd`) template the *entire* config and own the sole listener directive, so for them binding to `host:port` truly replaces the default — that doesn't carry over to squid's additive `conf.d` model.

roles/squid/templates/docker-compose.yml.j2
```diff
@@ -4,13 +4,18 @@ services:
     container_name: "{{ squid_container_name }}"
     restart: unless-stopped
     image: "{{ squid_image }}"
+{% if squid_network_mode == "host" %}
```
ideaship, line 7: `squid_network_mode` is compared only against `"host"`; any other value (a `"Host"` typo, a future `"none"`) silently renders bridge config here and in `osism.conf.j2`. Consider an `assert` early in `tasks/main.yml` restricting it to `bridge`/`host` so a misconfiguration fails loudly rather than silently changing the network posture. Non-blocking and consistent with `dnsdist`/`httpd`, but cheap insurance given this role's security stakes.

ideaship (changes requested):
### No test coverage for host mode

(Noted here rather than inline: `molecule/delegated/tests/squid.py` is not part of this diff.) The tests only exercise the default bridge path (`test_proxy_functionality`, line 53). There's no scenario asserting `squid_network_mode: host`, no check that the rendered compose has `network_mode: host` and no `ports:`, and no check that squid is *not* listening on `0.0.0.0:3128`. A host-mode case with the default `squid_port` would have caught the blocking finding at container-start time (squid exits FATAL). Even a render/parse-level assertion that no duplicate `http_port` results would flag it.

ideaship (approved):

PR #2115, opened 2026-06-17
Add optional host networking mode for squid

### LINKED ISSUE #2113: Run squid with host networking

fzakfeld:
Fixes https://github.com/osism/ansible-collection-services/issues/2113

### DIFF
--- a/roles/squid/defaults/main.yml
+++ b/roles/squid/defaults/main.yml
@@ -29,8 +29,9 @@ squid_service_name: "docker-compose@squid"
 
 squid_host: 127.0.0.1
 squid_port: 3128
+squid_host_network: false
 
-squid_tag: 6.1-23.10_beta  # don't get fooled by "beta". this is actually production ready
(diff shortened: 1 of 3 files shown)


berendt:
Thanks for this, @fzakfeld! Closing in favour of #2114, which implements the same feature but names the option `squid_network_mode` ("bridge"/"host") to stay consistent with the existing `httpd_network_mode` and `dnsdist_network_mode` options. It carries the same `http_port` host-binding you added here. Both close #2113.

PR #2147, opened 2026-07-21
frr: make maximum-paths configurable

### LINKED ISSUE #2122: FRR: bump maximum-paths setting or make it configurable
When FRR is configured to use a maximum of 2 paths towards other peers (the current default), but 4 connections and 4 paths exist, FRR picks 2, which might not be the same the switch picks when sending traffic back. This breaks the connections. For this setup, we need to allow setting `maximum-paths 4` to use all available paths. Either add an option for this, change the default to 4 (or maybe even higher?), or drop the setting completely, which should result in the number of paths to be unlimited (need to confirm this option, it might be possible that the default in FRR is to only allow 1 path, at least in some szenarios).

berendt:
The generic templates (`frr_leaf`, `frr_loadbalancer` and `frr_loadbalancer_external_uplink`) hardcoded `maximum-paths 2` per BGP address family. With four uplinks FRR then only installs two of the four ECMP paths, which do not necessarily match the paths the switches pick for the return traffic, breaking connections.

- Drop the hardcoded setting from the generic templates. Without an explicit `maximum-paths`, FRR falls back to its built-in default of `multipath_num` (256 in both the Debian and RedHat family packages, independent of the `frr defaults traditional` profile), so all available paths are used.
- The new `frr_maximum_paths` variable (`0` by default) renders the line again when a deployment wants to limit the number of paths. The templates check `| int > 0`, so a value of `0` and non-integer values do not write the statement.
- The yrzn templates keep their previous default of two paths; the hardcoded value in `frr_yrzn001.conf.j2` moves into the new `frr_yrzn001_maximum_paths` variable so it can be overridden like its metalbox and network counterparts.

Fixes #2122

### DIFF
--- a/roles/frr/defaults/main.yml
+++ b/roles/frr/defaults/main.yml
@@ -56,6 +56,14 @@ frr_vrrp_groups: []
 
 frr_incoming_table_external: 254
 
+# Maximum number of ECMP paths installed per BGP address family
+# (maximum-paths). With 0 no maximum-paths statement is written, so the
+# FRR built-in default applies (256 in the Debian and RedHat family
+# packages), which means all available paths are used. Set a value
+# greater than 0 only to limit the number of paths. Not used by the
+# yrzn templates.
+frr_maximum_paths: 0
+
 frr_local_pref:
   - 1000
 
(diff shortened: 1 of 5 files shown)


roles/frr/defaults/main.yml
```diff
@@ -56,6 +56,13 @@ frr_vrrp_groups: []
 
 frr_incoming_table_external: 254
 
+# Maximum number of ECMP paths installed per BGP address family
+# (maximum-paths). Unset by default, so the FRR built-in default
+# applies (256 in the Debian and RedHat family packages), which means
+# all available paths are used. Set a value only to limit the number
+# of paths. Not used by the yrzn templates.
+frr_maximum_paths: null
```
osfrickler: does this really evaluate to `false` in the template? maybe better use `0` as default and then check `| int > 0` in the template to also avoid setting non-integer values?

osfrickler (approved):

PR #2161, opened 2026-08-26
frr: add numbered BGP neighbors and prefix based bgp_out clauses

kingspeedy95:
The leaf template can only express the interface based unnumbered uplinks in `frr_uplinks`, and `bgp_out` can only ever export connected routes. Together that makes one common case unreachable from the configuration repository: a peer that is not on a directly connected link and that announces prefixes this router is supposed to carry into the fabric, for example a route injector or an OpenStack `neutron-dynamic-routing` agent.

Both halves are needed, and the second one is easy to miss.

### `frr_bgp_neighbors`

Numbered neighbors alongside the uplinks. Only `address` and `remote_as` are required; `description`, `password`, `passive`, `ebgp_multihop`, `update_source`, `bfd`, per family activation and per family route maps are optional.

```yaml
frr_bgp_neighbors:
  - address: 2001:db8::1
    remote_as: 65001
    description: route injector
    passive: true
    ebgp_multihop: 8
    activate_ipv4: false
    activate_ipv6: true
```

`activate_ipv4` defaults to true and `activate_ipv6` to false. That is FRR's own behaviour rather than a choice: `frr defaults traditional` implies `bgp default ipv4-unicast`, so a neighbor is already activated in the IPv4 address family and has to be switched off explicitly. A peer carrying only IPv6 therefore renders `no neighbor <address> activate` in IPv4, without which both sides negotiate an address family neither has anything to put in. `route_map_out` defaults to `bgp_out`, the same policy the uplinks already use.

### `frr_bgp_out_prefixes`

An interface clause cannot match a route learned from a peer at all. `route_match_interface()` compares the named interface against `attr->nh_ifindex`, which zebra sets when it redistributes a connected route and which is unset on a path received from a neighbor.

The failure is silent and takes a while to find: the session comes up, the route arrives, the kernel installs it, and nothing leaves the router.

```yaml
frr_bgp_out_prefixes:
  - prefix: 2001:db8:1::/48
    le: 64
```

Entries are split by family on the address itself and rendered as a prefix list plus a trailing `bgp_out` clause, numbered after the interface clauses so both lists can grow without colliding. A prefix list rather than `match peer`, so the export is bounded to the address space the operator designated instead of to whatever a peer happens to send.

### Verification

Rendered the template from `main` and this branch with identical variables. **With both new variables left at their defaults the output is byte for byte identical**, so nothing changes for anyone who does not set them.

The generated syntax was also entered into a running FRR 8.4.4 by hand before this was written, and the resulting session, route acceptance and re-export behaved as intended on real hardware.

`yamllint` with the repository configuration and `ansible-lint` at the production profile are clean on `roles/frr`.

### Note on test coverage

`frr_leaf.conf.j2` has no CI coverage today. `molecule/delegated/vars/frr.yml` sets `frr_type: test`, which renders `frr_test.conf.j2`, a static file with no variables in it, so the molecule job that the Zuul file matches to `roles/frr/**` never exercises this template. That is true of the whole template rather than only of this change, and giving the scenario real coverage looks like a separate change. Happy to follow up with one if that is wanted.


osfrickler (commented):
to be honest I don't like the idea of placing more complexity into the default template. in my idea it should only contain features that are to be used on most nodes in most deployments (I agree that this classification is a bit vague).

so at a minimum I think this should better be added into an extra template, maybe it would also better be kept local to your deployment, I think we may need to discuss a policy for this internally and document it properly

I'm also not sure yet about the specific use case you are referring to, which is connecting the neutron-dynamic-routing agent to the outside world, I'll need to do some local testing for that

kingspeedy95:
Thanks for the steer, it is out of the default template now.

The numbered neighbors and the prefix based `bgp_out` clauses live in their own template, `roles/frr/templates/frr_leaf_external_peers.conf.j2`, selected with `frr_type: leaf_external_peers`. That is the same shape `frr_loadbalancer_external_uplink` already has next to `frr_loadbalancer`, so it needs no new mechanism in the role, and `frr_leaf.conf.j2` is byte for byte the file on main again.

Verified by rendering all three cases with the same variables: the leaf template is unchanged, the new template with both variables at their defaults produces exactly what the leaf template produces, and setting them changes only the new neighbor, prefix list and route-map lines. So nobody sees anything until they opt in through `frr_type`.

On keeping it local to our deployment, I would rather not, and not because it is convenient for us. Nothing in it is specific to our fabric. Any leaf that peers with a BGP speaker which is not on a directly connected link needs both halves, and the second half is the one that costs people an afternoon: an interface clause in `bgp_out` can never match a route learned from a peer, because `route_match_interface()` compares against the ifindex zebra sets on a redistributed connected route and that is unset on a received path. The session comes up, the route arrives, the kernel installs it, and nothing leaves the router. A deployment that runs into that has no way to express the fix from its configuration repository today.

On the policy you want to discuss internally: a variant template keeps the default one clean whatever you decide, so I hope this does not have to wait for it.

On your local testing of the neutron-dynamic-routing case, tell me what would help. I can post a minimal `frr_bgp_neighbors` and `frr_bgp_out_prefixes` pair with the rendered `frr.conf` next to it, or the agent side configuration that produced it.

PR #2165, opened 2026-09-08
manager: support an HTTP proxy for the manager services

berendt:
Closes the manager side of osism/issues#1432.

The manager services use container networking. Where the outbound traffic of the manager network cannot leave the node directly, an HTTP proxy is the way out — but there was no way to configure one, and the workaround from the issue cannot work:

* `all.env` is rendered from `env/all.env.j2`, so a hand-written `HTTP_PROXY` in `/opt/manager/configuration/all.env` is overwritten by the next `osism apply manager`.
* `manager_environment_extra`, the only extension point today, ends up in `ansible.env` and therefore reaches `osism-ansible`, the other ansible services and `osismclient` — but not the `openstack` worker, which is the service that runs `openstack-image-manager`.

## What this changes

* New variables in the `manager` role, named after the pattern of the `docker` role: `manager_configure_proxy` (default `false`), `manager_proxy_http`, `manager_proxy_https`, `manager_proxy_no_proxy_default` / `_extra` / `manager_proxy_no_proxy`.
* `all.env` renders `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` in upper **and** lower case. Both spellings are deliberate: `requests` and `urllib` read either of them, `aria2c` — which `openstack-image-manager` uses to prefetch images — reads only the lowercase names.
* `NO_PROXY` is assembled from the service names of the manager, the manager network (plus the IPv6 network when enabled), the names from `ansible_services`, and `manager_proxy_no_proxy_extra` for internal endpoints outside of the manager network. Without it every internal call — netbox, api, vault, the internal OpenStack API — would be sent to the proxy.
* `all.env` is now read by every service that runs an OSISM image. The `env_file` block of `inventory_reconciler`, `api` and `beat` sat inside `{% if enable_netbox %}`, `watchdog` and `flower` had none at all, so without the netbox integration those services received neither the schedules nor the proxy. `netbox.env`, `inventory-reconciler.env` and `openstack.env` stay conditional, they are only rendered with netbox respectively celery enabled. The services that do not run an OSISM image (`redis`, `mariadb`, `ara-server`, `vault`, `frontend`) deliberately keep their own environment.

## Configuration

```yaml
manager_configure_proxy: true
manager_proxy_http: "http://10.0.0.10:3128"
manager_proxy_no_proxy_extra:
  - 192.168.16.0/20
  - api-int.example.com
```

## Testing

The templates were rendered with `ansible-playbook` for netbox on/off, celery on/off, proxy on/off and with an IPv6 manager network. The result parses as YAML in every combination and `all.env` is in the `env_file` list of every OSISM service. `yamllint` with the repository configuration and `flake8` are clean.

The molecule test for the manager role covers `all.env` from the converge as the disabled branch, the fixture keeps `manager_configure_proxy` off. The enabled branch is covered by `test_manager_proxy_environment`, which renders `env/all.env.j2` through Ansible with the role defaults and the molecule fixture as extra vars and asserts on the rendered values: the proxy URL in both spellings, `HTTPS_PROXY` following `manager_proxy_http`, and `NO_PROXY` carrying the manager network, the IPv6 network only when it is enabled, service names from `manager_proxy_no_proxy_default` and from `ansible_services`, and the entries of `manager_proxy_no_proxy_extra`. `test_docker_compose` asserts over the rendered compose file that every service running an OSISM image reads `all.env`, including `openstack`, `conductor`, `netbox` and `listener`.

## Not covered here

* `openstack-image-manager` is started with an empty environment by `python-osism`, so the proxy stops at that boundary until osism/python-osism#2682 is fixed.
* Images imported with `web-download` are downloaded by Glance itself, not by the manager. Those need the proxy on the kolla side, or the prefetch path.

### DIFF
--- a/molecule/delegated/tests/manager.py
+++ b/molecule/delegated/tests/manager.py
@@ -6,6 +6,47 @@ from packaging.version import Version
 
 testinfra_runner, testinfra_hosts = get_ansible()
 
+MANAGER_DEFAULTS = "../../roles/manager/defaults/main.yml"
+MANAGER_FIXTURE = "../../molecule/delegated/vars/manager.yml"
+ALL_ENV_TEMPLATE = "../../roles/manager/templates/env/all.env.j2"
+
+PROXY_VARIABLES = [
+    "HTTP_PROXY",
+    "http_proxy",
+    "HTTPS_PROXY",
+    "https_proxy",
+    "NO_PROXY",
+    "no_proxy",
+]
+
+
+def render_all_env(host, **variables):
+    """Render env/all.env.j2 of the manager role with the given variables.
+
+    The role defaults and the molecule fixture are passed to Ansible as extra
+    vars, so the templated defaults, above all manager_proxy_no_proxy, are
+    evaluated by Ansible in the same way as in the role. The given variables
+    take precedence over both. Returns the rendered file as a dictionary.
+    """
+    variables = {
+        **host.ansible("include_vars", MANAGER_DEFAULTS)["ansible_facts"],
+        **host.ansible("include_vars", MANAGER_FIXTURE)["ansible_facts"],
+        **variables,
+    }
+    path = host.ansible("tempfile", "suffix=.env", check=False)["path"]
+    try:
+        host.ansible(
+            "template",
+            f"src={ALL_ENV_TEMPLATE} dest={path}",
+            check=False,
+            extra_vars=variables,
+        )
+        content = host.file(path).content_string
+    finally:
+        host.ansible("file", f"path={path} state=absent", check=False)
+
+    return dict(line.split("=", 1) for line in content.splitlines() if line)
+
 
 # testing config.yml tasks
 def test_manager_config(host):
@@ -45,20 +86,11 @@ def test_manager_config(host):
     assert all_env.group == get_variable(host, "operator_group")
     assert "INVENTORY_RECONCILER_SCHEDULE=" in all_env.content_string
 
-    proxy_variables = [
-        "HTTP_PROXY",
-        "http_proxy",
-        "HTTPS_PROXY",
-        "https_proxy",
-        "NO_PROXY",
-        "no_proxy",
-    ]
-    if get_variable(host, "manager_configure_proxy"):
-        for variable in proxy_variables:
-            assert f"{variable}=" in all_env.content_string
-    else:
-        for variable in proxy_variables:
-            assert f"{variable}=" not in all_env.content_string
+    # NOTE: The molecule fixture keeps the proxy disabled, this is the
+    #       disabled branch of all.env.j2. The enabled branch is covered by
+    #       test_manager_proxy_environment on a rendering of the template.
+    for variable in PROXY_VARIABLES:
+        assert f"{variable}=" not in all_env.content_string
 
     # config-ara
     if host.file(get_variable(host, "enable_ara")) == "true":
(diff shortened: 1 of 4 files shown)


molecule/delegated/tests/manager.py
```diff
@@ -36,6 +37,29 @@ def test_manager_config(host):
...
+
+    proxy_variables = [
+        "HTTP_PROXY",
+        "http_proxy",
+        "HTTPS_PROXY",
+        "https_proxy",
+        "NO_PROXY",
+        "no_proxy",
+    ]
+    if get_variable(host, "manager_configure_proxy"):
```
ideaship: Please add coverage that actually exercises the enabled branch. Nothing under `molecule/` sets `manager_configure_proxy`, and the role default is `false`, so in CI only the `else` branch runs, the one asserting the proxy variables are *absent*. The branch that checks what this PR adds never executes, which also makes the description's "both proxy branches" claim stronger than what runs.

I am not asking for a specific mechanism. Setting the variable in `molecule/delegated/vars/manager.yml` would be four lines, but it puts an unreachable proxy into the shared fixture for every manager service for the rest of the scenario, so a dedicated template-rendering check may well be the better shape, your call. What I would like either way is that the enabled path is asserted on rendered values, not just on the presence of the names: the manager network, a couple of representative service names, and an operator-supplied `manager_proxy_no_proxy_extra` entry. `manager_proxy_no_proxy` is the expression most likely to break silently under a later refactor, and a wrong `NO_PROXY` sends internal traffic to the proxy instead of failing loudly. Keeping the disabled branch covered as well would be good.

molecule/delegated/tests/manager.py
```diff
@@ -229,6 +253,25 @@ def test_docker_compose(host):
                 rf'container_name: "{container}"', f.content_string
             ), f"Container name '{container}' not found in docker-compose.yml"
 
+    # NOTE: all.env carries the proxy configuration, every service of the
+    #       manager has to read it, also without the netbox integration.
+    all_env = f"{get_variable(host, 'manager_configuration_directory')}/all.env"
+    services = yaml.safe_load(f.content_string)["services"]
+    for name in [
```
ideaship: Please add `openstack` to this list, and `conductor`, `netbox` and `listener` with it. `openstack` is the service this change exists for, it runs `openstack-image-manager`, and its `env_file` block is where the proxy has to survive.

All four already read `all.env`, so nothing is broken now, which is exactly why they are worth asserting: the comment above the loop says every service of the manager has to read it, and the four workers are the gap in that. A future edit to the `openstack` block would pass this test as it stands. The `if name not in services: continue` guard makes the addition free.

roles/manager/templates/env/all.env.j2
```diff
@@ -1,2 +1,14 @@
...
+{% if manager_configure_proxy | bool %}
+{# NOTE: Both spellings are set on purpose. requests, urllib and the Go
+         based tools accept either of them, aria2c, which is used by the
+         openstack-image-manager to prefetch images, only reads the
+         lowercase names. #}
+HTTP_PROXY={{ manager_proxy_http }}
+http_proxy={{ manager_proxy_http }}
+HTTPS_PROXY={{ manager_proxy_https }}
+https_proxy={{ manager_proxy_https }}
+NO_PROXY={{ manager_proxy_no_proxy | join(',') }}
```
ideaship: We want `REQUESTS_CA_BUNDLE` rendered for the manager services too. This does not have to hold up the proxy support, a follow-up PR is fine, but please don't leave it unowned, because the two features correlate.

The counterpart's allowlist carries the CA variables alongside the proxy names, and after this PR the proxy half has a delivery path while the CA half has none: `manager_environment_extra` reaches only `ansible.env`, which the `openstack` worker does not read, and its commented example in `defaults/main.yml:372` is literally `REQUESTS_CA_BUNDLE: /etc/ssl/certs/ca-certificates.crt`.

The case is narrower than "any proxy": it is a proxy that intercepts TLS with a private CA the client does not already trust. There it bites, because the manager services mount the host bundle into the container while `requests` uses certifi unless `REQUESTS_CA_BUNDLE` or `CURL_CA_BUNDLE` is set, so the operator gets working reachability and then a certificate failure, inside a Celery task whose output is not usually read back.

`REQUESTS_CA_BUNDLE` alone is enough for the `requests`-based path, which is the one that breaks here; the four names in the counterpart's allowlist are alternative mechanisms for different consumers, not four settings to configure. The value must be the container path `/etc/ssl/certs/ca-certificates.crt`, not `_manager_ca_certificates_file`, which is the host-side mount source and differs on Red Hat. `roles/openstackclient/templates/docker-compose.yml.j2:21` and `roles/substation/templates/docker-compose.yml.j2:24` already do exactly this.

roles/manager/defaults/main.yml
```diff
@@ -371,6 +371,48 @@ manager_wrapper_scripts:
...
+  - netbox
+  - openstack
+  - osismclient
+  - redis
+  - vault
+  - watchdog
+# NOTE: Internal endpoints outside of the manager network belong here, for
+#       example the internal OpenStack API, an internally reachable netbox or
+#       the networks of the managed nodes.
+manager_proxy_no_proxy_extra: []
```
ideaship: Please add a line to the manager documentation next to `manager_configure_proxy`. `manager_proxy_no_proxy_extra` is empty by default while the endpoints most likely to need it, the internal OpenStack API, the managed node networks, a netbox reachable outside the manager network, may sit outside what the defaults cover. The NOTE here says that, so the code is fine; the question is whether an operator following the deployment docs rather than reading `defaults/main.yml` finds out before the first failed internal call. The example block in the PR description is already the right content.

molecule/delegated/tests/manager.py
ideaship: Please add coverage that actually exercises the enabled branch. Nothing under `molecule/` sets `manager_configure_proxy`, and the role default is `false`, so in CI only the `else` branch runs, the one asserting the proxy variables are *absent*. The branch that checks what this PR adds never executes, which also makes the description's "both proxy branches" claim stronger than what runs.

I am not asking for a specific mechanism. Setting the variable in `molecule/delegated/vars/manager.yml` would be four lines, but it puts an unreachable proxy into the shared fixture for every manager service for the rest of the scenario, so a dedicated template-rendering check may well be the better shape, your call. What I would like either way is that the enabled path is asserted on rendered values, not just on the presence of the names: the manager network, a couple of representative service names, and an operator-supplied `manager_proxy_no_proxy_extra` entry. `manager_proxy_no_proxy` is the expression most likely to break silently under a later refactor, and a wrong `NO_PROXY` sends internal traffic to the proxy instead of failing loudly. Keeping the disabled branch covered as well would be good.

molecule/delegated/tests/manager.py
ideaship, line 340: Please add `openstack` to this list, and `conductor`, `netbox` and `listener` with it. `openstack` is the service this change exists for, it runs `openstack-image-manager`, and its `env_file` block is where the proxy has to survive.

All four already read `all.env`, so nothing is broken now, which is exactly why they are worth asserting: the comment above the loop says every service of the manager has to read it, and the four workers are the gap in that. A future edit to the `openstack` block would pass this test as it stands. The `if name not in services: continue` guard makes the addition free.

ideaship (changes requested):

roles/manager/templates/docker-compose.yml.j2
```diff
@@ -42,8 +42,10 @@ services:
 {%   if netbox_secondaries | length > 0 %}
       - NETBOX_SECONDARIES
 {%  endif %}
```
osfrickler: you could fix the indentation here to reduce the confusion about whether the added `endif` is correct or not. same thing twice below

osfrickler (commented):
the change to the role lgmt, just some nits. I do agree with the testing comments

osfrickler (approved):

PR #2170, opened 2026-09-15
httpd: serve the same content on the TLS virtual host

kingspeedy95:
## The gap

The TLS virtual host carries only the document root and the `/ironic` alias. The plain one additionally carries the `/sonic` alias with its ONIE serial number rewrite, and the API proxy for `/v1/`.

So enabling `httpd_ssl_enable` produces a virtual host that silently serves a subset. SONiC ZTP and the API stay reachable over plain HTTP only, and nothing in the role expresses that as a decision.

## The change

Both bodies become one macro, called from both virtual hosts, so they cannot drift again.

Nothing new is exposed. Every block keeps the flag it already had, `httpd_sonic_ztp_enable` and `httpd_ironic_enable`, and the same files were already served by the same Apache on port 80.

The alias and the rewrite deliberately travel together. Serving `/sonic` over TLS with only the alias would hand every switch the generic ONIE installer and silently lose the per device pinning that `httpd_sonic_ztp_onie_installer_per_device` exists for.

The TLS virtual host also carried its own `ErrorLog` and `CustomLog`, byte identical to the global directives at the bottom of the same file, including the `206` filter. Virtual hosts inherit the main server's logging unless they override it, so those two lines changed nothing; they are removed rather than copied to the plain host, which is what makes one shared body possible.

## Verification

The template was rendered before and after across all sixteen combinations of `httpd_network_mode`, `httpd_ssl_enable`, `httpd_sonic_ztp_enable` and `httpd_ironic_enable`:

* the plain virtual host renders **byte identical** in all sixteen
* with `httpd_ssl_enable` false the **whole file** renders byte identical, so deployments not using TLS see no change at all
* with it true the TLS host gains `/sonic`, the rewrite and the proxy exactly when their own flags are set

## Found while reading, deliberately not changed

The `/v1/` proxy sits inside the `httpd_sonic_ztp_enable` block, so the API proxy depends on a SONiC flag. That looks unintended, but it is pre-existing behaviour and moving it would change what deployments get today. Happy to follow up separately if you want it split out.

`Options Indexes` is set under both aliases, so `/sonic/` and `/ironic/` are browsable, now over TLS as well as plain HTTP. Unchanged here, mentioned because this PR widens where it applies.

## Why

Found while moving an Ironic virtual media image endpoint to HTTPS, where some BMC firmware refuses plain HTTP outright. That part works; ZTP over HTTPS is what this unblocks.


roles/httpd/templates/httpd.conf.j2
```diff
@@ -133,28 +142,7 @@ LoadModule ssl_module modules/mod_ssl.so
...
-    Alias /ironic /var/lib/ironic/httpboot
-    <Directory "/var/lib/ironic/httpboot">
-        Options Indexes FollowSymLinks
-        AllowOverride All
-        Require all granted
-    </Directory>
-{% endif %}
-
-    ErrorLog /proc/self/fd/2
-{% if httpd_log_ignore_partial_content | bool %}
```
osfrickler, line 153: why are you dropping this? it isn't even mentioned in the commit message

roles/httpd/templates/httpd.conf.j2
```diff
@@ -74,11 +74,13 @@ LoadModule ssl_module modules/mod_ssl.so
...
-{% if httpd_network_mode == "host" %}
-<VirtualHost {{ httpd_host | ansible.utils.ipwrap }}:{{ httpd_port }}>
-{% else %}
-<VirtualHost 0.0.0.0:80>
-{% endif %}
+{#
+  Body shared by the plain and the TLS virtual host. Defined once so the two
+  cannot drift: before this existed the TLS host carried only the document root
+  and /ironic, so SONiC ZTP and the API proxy were reachable over plain HTTP
+  only, with nothing in the role saying that was intended.
```
osfrickler: the history can be seen in git, it doesn't need to be mentioned in this comment. so just the first sentence is enough IMO

osfrickler (changes requested):

roles/httpd/templates/httpd.conf.j2
kingspeedy95: Restored. The TLS virtual host keeps its own `ErrorLog` and `CustomLog` unchanged, and the shared body is now emitted in front of them, so this virtual host only gains content and loses nothing.

The removal was meant as a cleanup and argued in the commit message, the two lines being byte identical to the global directives at the bottom of the file, which a virtual host inherits when it does not override them. That is a separate question from serving the same content on both hosts though, so it does not belong in this change and is gone.

roles/httpd/templates/httpd.conf.j2
kingspeedy95: Cut to one sentence, thanks.

We took the point beyond this file: a comment saying what the thing is for, with the history left to git and the reasoning in the commit message, is now a written rule on our side.

osfrickler (approved):

PR #2171, opened 2026-09-16
Add snmp_exporter role

fzakfeld:
AI-assisted: Claude:claude-sonnet-5


ideaship (changes requested):
https://zuul.services.osism.tech/t/osism/buildset/bac78be2aa1a4208b10791501ac1bb22

The Zuul checks failed because there is no README.md in src/github.com/osism/ansible-collection-services/roles/snmp_exporter.

We are trying to add consistent commit messages along with our commits, see
https://osism.tech/docs/guides/developer-guide/style-guide#commit-messages.

roles/snmp_exporter/defaults/main.yml
```diff
@@ -0,0 +1,46 @@
...
+snmp_exporter_docker_compose_directory: /opt/snmp-exporter
+
+snmp_exporter_container_name: snmp-exporter
+snmp_exporter_service_name: "docker-compose@snmp-exporter"
+
+snmp_exporter_host: 0.0.0.0
+snmp_exporter_port: 9116
+snmp_exporter_port_container: "{{ snmp_exporter_port }}"
+
+snmp_exporter_flags_defaults:
```
ideaship: `snmp_exporter_host`, `snmp_exporter_port` and `snmp_exporter_port_container` are defined but never used. Nothing in the role passes them to the exporter, and the container runs with `network_mode: host`, so snmp_exporter falls back to its built-in `--web.listen-address=:9116` on every interface. Setting `snmp_exporter_host: 127.0.0.1` or `snmp_exporter_port: 9200` has no effect.

Please add the flag here, the way `frr` does (`roles/frr/templates/prometheus-frr-exporter.j2:4`):

```yaml
snmp_exporter_flags_defaults:
  - "--config.file=/etc/snmp_exporter/snmp.yml"
  - "--web.listen-address={{ snmp_exporter_host | ansible.utils.ipwrap }}:{{ snmp_exporter_port }}"
```

`snmp_exporter_port_container` can go. With host networking there is no container port to map.

roles/snmp_exporter/defaults/main.yml
```diff
@@ -0,0 +1,46 @@
...
+##########################
+# snmp_exporter
+
+snmp_exporter_configuration_directory: /opt/snmp-exporter/configuration
+snmp_exporter_docker_compose_directory: /opt/snmp-exporter
+
+snmp_exporter_container_name: snmp-exporter
+snmp_exporter_service_name: "docker-compose@snmp-exporter"
+
+snmp_exporter_host: 0.0.0.0
```
ideaship: Once the flag is wired up, please default this to `127.0.0.1`. That's what the closest siblings do: `scaphandre`, which this role is modelled on, and `frr_exporter`, the other exporter, both default to `127.0.0.1`.

roles/snmp_exporter/defaults/main.yml
```diff
@@ -0,0 +1,46 @@
...
+snmp_exporter_port: 9116
+snmp_exporter_port_container: "{{ snmp_exporter_port }}"
+
+snmp_exporter_flags_defaults:
+  - "--config.file=/etc/snmp_exporter/snmp.yml"
+snmp_exporter_flags_extra: []
+snmp_exporter_flags: "{{ snmp_exporter_flags_defaults + snmp_exporter_flags_extra }}"
+
+snmp_exporter_tag: "v0.27.0"
```
ideaship: Nit: `scaphandre` and `kepler` let renovate track this fallback tag. Doing the same here needs the `# renovate: datasource=docker depName=index.docker.io/prom/snmp-exporter` annotation above the tag, single quotes (`snmp_exporter_tag: 'v0.27.0'`, because the regex in `.github/renovate.json` requires them), and `^roles\/snmp_exporter\/defaults\/main.yml$` in that file's `fileMatch` list. Please also check that `osism/release` gets a matching entry, since that is where the deployed version comes from.

roles/snmp_exporter/defaults/main.yml
```diff
@@ -0,0 +1,46 @@
...
+##########################
+# configuration
+
+configuration_directory: /opt/configuration
+
+# Optional custom snmp.yaml (MIB walk configuration). When a file exists at
+# this path on the control node it is copied to the target host and bind
+# mounted into the container, replacing the snmp.yml bundled with the image.
+# Store an override in the local Ansible environment at:
+#   {{ configuration_directory }}/environments/generic/files/snmp_exporter/snmp.yaml
```
ideaship: The comment says `environments/generic/...`, but the default below uses `environments/infrastructure/...`. I think `environments/monitoring/files/snmp_exporter/snmp.yaml` would be better than either. Roles take their files from the environment whose playbook deploys them (`frr` from `generic`, `netbox` from `infrastructure`), and the other exporters (`scaphandre`, `kepler`, `gnmic`) are deployed from `playbooks/monitoring/` in `osism/ansible-playbooks`. So I'd expect this path to match wherever the snmp_exporter playbook goes. Could you open that playbook PR? Being able to review both together would help.

molecule/delegated/tests/snmp_exporter.py
```diff
@@ -0,0 +1,50 @@
...
+    assert (
+        docker_compose_file.mode == 0o640
+    ), "docker-compose.yml should have 0640 permissions"
+
+
+def test_snmp_exporter_service_running(host):
+    snmp_exporter_service_name = get_variable(host, "snmp_exporter_service_name")
+    service = host.service(snmp_exporter_service_name)
+    assert service.is_enabled
+    assert service.is_running
```
ideaship: The tests currently check directories, the compose file and that the service is running, so they would not have caught the missing flag. Please add a bind test modelled on `test_exporter_defaultsfile` and `test_exporter_listening` in `molecule/delegated/tests/frr.py:112-164`:

- assert that the rendered `docker-compose.yml` contains `--web.listen-address=<host>:<port>`, and
- assert that every address `ss -ltnH 'sport = :<port>'` reports equals `snmp_exporter_host:snmp_exporter_port`.

The second check needs to use `ss`, not just `host.socket(...).is_listening`. As the comment in `frr.py` explains, `is_listening` against `127.0.0.1` also succeeds when the exporter is bound to a wildcard address, and snmp_exporter binds `[::]` (dual-stack), so a "not `0.0.0.0`" check would not catch it either. With the `ss` version, the test fails against the current code and passes once the flag is added.

ideaship (changes requested):

